PR #24, opened 2022-03-13
First Version of NavBar

Chenz-7:
Not sure if the color of the logo is going to stay like that or be changed, and logo links to index.tsx.


eboysen:
This needs to be moved into a feature branch so that the code can be viewed on a local machine, would recommend finding a way to prevent code duplication, if you need help please reach out

PR #29, opened 2022-03-24
Implement an expandable search grid UI

iamwood:
## Overview

Replaces placeholder search grid with proper ExpandableSearchGrid component, as well as adding a secondary text value to the SearchTermCard.

## What Changed

ExpandableSearchGrid now stores options selected as the whole Film object (to be replaced with the classes data form). Some visual changes to the grid, term cards, and autocomplete dropdown.


eboysen (approved):
Looks good, make sure to add comments before final push

PR #31, opened 2022-03-25
Implement Autocomplete Option API

### LINKED ISSUE #30: [Feature Request] Autocomplete Options API Implementation
# Overview

The autocomplete options should be taken from an API suggesting potential matching courses and professors for those courses.

When the user types in the input, a call is made with the text and the options are updated to those.

# Proposal

The autocomplete API needs to be implemented in the two instances of the search bars.

iamwood:
## Overview

Resolves #30 

The autocomplete API is now being used by the search bars, so the options presented are the relevant search suggestions that will be able to pull data.

## What Changed

The autocomplete API is called using Axios get request. Instances of Film data type have been replaced with Suggestion.


WillieCubed (commented):
Prof. Cole gives a 👍

Issue #28, opened 2022-03-23
[Feature Request] Rebuild home page


eboysen:
# Overview

The home page has gone through a redesign in hopes of increasing the readability and cohesiveness of the design language. There will be a couple of simple changes made to it most of which are some simple reformatting of the divs surrounding the search, logo and title. 

Recommend that you take a look at issue #27 as well if working on this feature branch.

Functionality will remain the same, only the html template and formatting need be adjusted.

# Proposal
The overlapping cards are being replaced with a single semi-transparent card, this should be a relatively simple fix. Simply edit the tsx file's return to adjust the html template so that it reflects the design better. Additionally, there are some spacing issues on mobile currently that should be adjusted at the same  time to reduce the margin and padding when on smaller screens.

Here is an screenshot from the design docs
![image](https://user-images.githubusercontent.com/50427509/159760225-2afc7b9d-f6ad-4918-9935-0f0fd301cf33.png)

George-Zhang5246:
Maybe close this issue? I feel like the current homepage is better than the image.

iamwood:
I agree. Once we get design and product people in to talk with us, I think we'll probably have redesign work all over, but we'll create a new issue for that once we get there.

Issue #27, opened 2022-03-21
[Bug]: Splash Page Components Don't Scale Properly


shoaibhuq:
### What did you expect to happen?

When zooming in and out of the website outside the default state, all the components should scale relative to the percentage of zoom the user selects. 

### What happened?

- The vector in the background does not stretch properly
- The cards behind the logo and the search bar don't scale down with the zoom percentage
- The logo does not scale down properly

Worst Case (Max Zoom Out): 
<img width="1721" alt="image" src="https://user-images.githubusercontent.com/44373355/159348612-9d7f7398-3feb-4365-9f0b-0711bfa1fa65.png">

### Version

Beta 0.1.0

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari, Microsoft Edge

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

shoaibhuq:
This issue is mostly fixed.
Design is subject to change so this may not be relevant anymore.

PR #34, opened 2022-04-03
chore: Homepage Rework (Issues #27 and #28)

penczak:
Fixed bug of background SVG waves zooming out and implemented design changes.
TODO: Change logo on homepage.

## Overview

Addresses Issues #27 and #28. 

I implemented the new desired design for the homepage and fixed the zoom bug. 

## What Changed

pages/index.tsx: Changed the return to have only one card instead of two and to match the Figma.
components/common/waveSVG.tsx and waveSVG2.tsx: The zoom bug was due to something here. I just changed the width and height properties of the <svg> to be 100% rather than pixel values. 
Figma:
image

My version:
image

The only thing that still needs to be changed is the logo, but Tayisiya 

## Other Notes

In line 38 of index.tsx, I added a <br/> to get the spacing right on the logo, but this doesn't seem like a good practice. I couldn't get normal margins to work, so I did this as a work around.

### DIFF
--- a/components/common/splashPageSearchBar.tsx
+++ b/components/common/splashPageSearchBar.tsx
@@ -134,7 +134,7 @@ export const SplashPageSearchBar = (props: SearchProps) => {
                 type="search"
                 id="mainSearch"
                 className="rounded-md border-primary-dark border-2 w-full h-12 pl-12 bg-white text-primary-dark placeholder-primary-light font-bold"
-                placeholder="CS1337..."
+                placeholder="Search section number, professor name, course number...."
               />
             </div>
           )}
--- a/components/common/splashPageSearchBar.tsx
+++ b/components/common/splashPageSearchBar.tsx
@@ -133,8 +133,8 @@ export const SplashPageSearchBar = (props: SearchProps) => {
                 {...params.inputProps}
                 type="search"
                 id="mainSearch"
-                className="rounded-md border-primary-dark border-2 w-full h-12 pl-12 bg-white text-primary-dark placeholder-primary-dark font-bold"
-                placeholder="Search section number, professor name, course number...."
+                className="rounded-md border-primary-dark border-2 w-full h-12 pl-12 bg-white text-primary-dark placeholder-primary-light font-bold"
+                placeholder="CS1337..."
               />
             </div>
           )}
--- a/pages/index.tsx
+++ b/pages/index.tsx
@@ -32,14 +32,11 @@ const Home: NextPage = () => {
         </div>
         <div className="flex justify-center content-center h-screen" style={{marginLeft: "10vw", marginRight: "10vw"}}>
           <div className="w-full relative m-auto">
-            <Card className="bg-light relative overflow-visible drop-shadow-lg rounded-xl bg-opacity-75 sm:w-3/4 xs:w-11/12 md:w-3/4" style={{marginTop: "1rem", marginBottom: "1rem", marginLeft: "auto", marginRight: "auto"}}>
-              <div className="m-auto md:w-1/6 sm:w-1/8 xs:w-1/8" style={{maxWidth: "50%", color: "black"}}>
-                <br/>
-                <br/>
-                <br/>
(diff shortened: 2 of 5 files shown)


components/common/splashPageSearchBar.tsx
```diff
@@ -133,8 +133,8 @@ export const SplashPageSearchBar = (props: SearchProps) => {
                 {...params.inputProps}
                 type="search"
                 id="mainSearch"
-                className="rounded-md border-primary-dark border-2 w-full h-12 pl-12 bg-white text-primary-dark placeholder-primary-dark font-bold"
-                placeholder="Search section number, professor name, course number...."
+                className="rounded-md border-primary-dark border-2 w-full h-12 pl-12 bg-white text-primary-dark placeholder-primary-light font-bold"
+                placeholder="CS1337..."
```
eboysen, line 137: This probably does not need to be changed to match the figma exactly, previous placeholder was fine too

pages/index.tsx
```diff
@@ -30,27 +30,24 @@ const Home: NextPage = () => {
...
-                  selectSearchValue={searchOptionChosen}
-                  value={value}
-                  setValue={setValue}
-                  disabled={false}
-                />
+        <div className="flex justify-center content-center h-screen">
+          <div className="w-full relative m-auto">
+            <Card className="bg-light relative overflow-visible drop-shadow-lg rounded-xl bg-opacity-75 m-auto sm:w-5/12 xs:w-11/12 md:w-3/5 ">
+              <div className='' style={{margin: "1rem"}}>
+                <div className="m-auto w-1/6" style={{margin: "1rem"}}>
```
eboysen, line 37: Try adding a color: "black" to the style tag on this div. The SVG should be set to be text color responsive, this will help prevent the bordered look of the logo.

eboysen (changes requested):
Regarding issue #27, try to get the logo to also scale with zooming, that seems to be the only component that is still not scaling properly. Also see other code comments regarding adjusting logo.

penczak:
Ok, I changed to logo color and changed the placeholder text back. I kept the "light" placeholder text because I think it looks much better, but if you want me to change that back to the dark, I can.

eboysen (approved):

eboysen:
Looks good on my end, clean up the mobile experience a little bit too. I'm gonna assign Shoaib to take a look real quick too since he was the original person to submit the bug report to see if it matches what he expected.

shoaibhuq:
Regarding Issue #27 

At 100% the font size for the title is much larger than what I desired in the Figma.
image

At 25% you can see clearly that the text label is not scaling currently
image



Regarding #28 

imageimage

There are a few issues between what is desired in the design and the current deployment:

- Background gradient is incorrect currently, it should go from colors: `7486CE` to `BCC9FD`
- The white backcard should be larger
- Title Label 
>Use font size for `Heading 1` 
- Textview box should be larger
>Font size in text view box should be `Body 1`

Refrence for Font Size
image

shoaibhuq (changes requested):
https://github.com/UTDNebula/athena/pull/34#issuecomment-1086965458

shoaibhuq (approved):

eboysen (changes requested):
Almost there, found one more thing in need of fixing. For devices where the screen height is smaller than the grey box, the box overflows and the background is cut off. This should be fixable based upon using a percentage of height rather than a static one for the box so that it stays on a single screen. If you want to replicate, select a mobile device in the responsive layouts and rotate it to landscape
image

penczak:
> Almost there, found one more thing in need of fixing. For devices where the screen height is smaller than the grey box, the box overflows and the background is cut off. 

Good catch.. I would not have noticed that. Should be fixed now.

penczak:
So to merge with the StyleGuidist branch, will I just close this PR and make another one after @eboysen  pushes the StyleGuidist directory changes?

eboysen:
> So to merge with the StyleGuidist branch, will I just close this PR and make another one after @eboysen pushes the StyleGuidist directory changes?

Yes, we likely will go about it like that, I will wait until a new branch and PR open up before closing anything out though

PR #38, opened 2022-04-23
Feat/styleguidist docs for graph components

penczak:
## Overview

rest of the project.

This PR adds interactive Styleguidist docs for all of our graph components. These will be useful to developers for quickly learning how these components work and what form their props should be in. 
This PR also does some refactors on the BoxGraph and RadialBarChart components. Previously, these two components required their props.series to be of unique formats, but now they accept the same { name: "", data: [] } format as all the other components. 

## What Changed

Screenshot of BoxGraph example. 
image
image

And a couple of screenshots of GraphChoice. 
image
image


eboysen (changes requested):
Everything looks pretty solid so far. Some of the features didn't work quite as expected. For example I would have liked to see the adding of Axis lablels to also expand the datasets to match the new number of labels
image
Also doesn't appear to be working on the Line Graph or Radar Chart at all
image
image


One of the small issues that I noticed is not necessarily within the scope of this PR however. The Box Graph seems to have the ability to overflow when it goes from 1 series up to 2. 
image

Other than that, it looks like great work. Let me know if you need anything!

penczak:
> For example I would have liked to see the adding of Axis lablels to also expand the datasets to match the new number of labels

I left it this way to show that you need to actually add more data when you add more xaxisLabels. I intentionally kept those separate. If you still think its a better design choice to have more data generated when you add labels, I could add that. 
> Also doesn't appear to be working on the Line Graph or Radar Chart at all

This is also because those graph types don't actually display any xaxisLabels beyond the number that matches the data it is given. I left the styleguidist like this to show those aspects of the graphs.

eboysen (approved):

eboysen:
That sounds like a good justification to me. Nice work on this, it looks great

PR #42, opened 2022-09-13
Changed references of "Athena" to "UTD-Trends"

Master7130:
## Overview

Partially resolves #39 

I replaced all occurrences of "Athena" to "UTD-Trends" except for the files in the styleguide folder.

## What Changed

The CONTRIBUTING.md file was updated so that new contributors won't be confused since the repo is named utd-trends while in the contribution file it's referred to as athena. I also updated the README.md, package.json and other files.

## Other Notes

There were some occurences of "athena" in files in the styleguide folder but I'm not exactly sure what the files in that folder were for so I didn't modify them.

PR #53, opened 2022-10-18
Changes Athena to UTD Trends

Seichro:
Changes Athena on homepage to UTD Trends from Rebuild Home Page #28


iamwood:
This was handled by a previous pull request #42.

PR #55, opened 2022-10-18
Replaced all instances of "Athena" with "UTD-Trends"

Aarian-A:
## Overview

Replaced all instances of "Athena" with "UTD-Trends" via the find and replace function.


shoaibhuq (approved):
Looks good!

PR #67, opened 2023-03-24
Feature/data api connection

iamwood:
Creates all of the data connections and professor cards for the initial launch of Trends


iamwood:
Initial launch features

PR #77, opened 2023-03-28
Yaxis formatting

TyHil:
## Overview

Use percents on main grades graph and GPAs with 2 decimal points on others.

## What Changed

Added a graph prop that passes a function with which to format y axis labels into a string.


iamwood (approved):
This is a good addition, looks great

PR #79, opened 2023-03-28
Consistency edits

George-Zhang5246:
## Overview

Basic edits to fix colors and use react components like Box and Typography instead of div and headers

## What Changed

ProfessorCard: Changed div to Box
CardContents: Made colors consistent, changed div to Box, changed h1 and h2 to Typography


iamwood (approved):
Good change, consistency is good

Issue #74, opened 2023-03-25
[Feature Request] Student aggregation total count display


TyHil:
# Overview

To better display that Trends is aggregating many sections across academic sessions, display the total number of students being aggregated in the expandable search grid.

# Proposal

Once data has been loaded from the API pass an array back to the expandable search grid containing the total number of students aggregated in each chosen data set (this is already calculated on the dashboard). Then, using the `secondaryText` parameter in the `searchTermCard.tsx` component, display this data to the user.

## Supplementary Information
Concept image (fake numbers):
![Screenshot 2023-03-24 211126](https://user-images.githubusercontent.com/63886162/227681597-f1f21088-8146-49a1-ba06-6f2e05eb9ce6.png)

PR #80, opened 2023-03-28
Feature/aggregation total count display

TyHil:
## Overview

Implements #74

Display total student grades being aggregated in the expandable search grid.

## What Changed

Moved the startingData generation and router from the dashboard to the expandable search grid so that it could take dynamic props. Then passed a changed prop for each data change (pull new data or change academic session range) for display.


iamwood (approved):
Solid change, ready to merge

Issue #72, opened 2023-03-25
[Feature Request] Autocomplete section suggestions


TyHil:
# Overview
Expand the autocomplete graph to suggest sections.

# Proposal

Increase the allowable formats the autocomplete graph suggests from (note spaces and that brackets represent optionality):
```
<prefix>[<number>| <number>[ <professorLast>|(<professorFirst> <professorLast>)]]
<professorLast>|(<professorFirst> <professorLast>)[ <prefix>[<number>| <number>]]
```
to:
```
<prefix>[<number>| <number>[.<section>][ <professorLast>|(<professorFirst> <professorLast>)]]
<professorLast>|(<professorFirst> <professorLast>)[ <prefix>[<number>| <number>[.<section>]]]
```

This change would be implemented in the `generateAutocompleteGraph.ts` file under the `addPrefixFirst` and `addProfFirst` functions.

TyHil:
Turns out we do not actually want sections in the autocomplete as they're just an ambiguous subset of the grades for a course. Since professors change for sections.

This does still however include exclusive sections like .HON and may end up including .HNx and .0Lx
This also optimizes the autocomplete system by implementing a radix tree and limiting the number of responses returned.

PR #82, opened 2023-03-30
Feature/autocomplete section suggestions

TyHil:
## Overview

Implements #72 

Add section data to autocomplete suggestion DAG.
Reformat DAG as a radix DAG to reduce file size.

## What Changed

`generateAutocompleteGraph.ts`
Added sections from preexisting data.
Radix tree generation with recursive node combining.

`autocomplete.ts`
Update autocomplete api to support nodes with more than one character.

## Other Notes

The graph JSON file is now 24MB.

### DIFF
--- a/components/common/ExpandableSearchGrid/expandableSearchGrid.tsx
+++ b/components/common/ExpandableSearchGrid/expandableSearchGrid.tsx
@@ -114,11 +114,11 @@ function searchQueryLabel(query: SearchQuery): string {
   if (query.number !== undefined) {
     result += ' ' + query.number;
   }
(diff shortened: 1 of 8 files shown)


TyHil:
Turns out we do not actually want sections in the autocomplete as they're just an ambiguous subset of the grades for a course. Since professors change for sections.

This does still however include exclusive sections like .HON and may end up including .HNx and .0Lx later, pending API support.
This also optimizes the autocomplete system by implementing a radix tree and limiting the number of responses returned.

iamwood (approved):
Looks good, I think we have a decent system for expanding the formats allowed by the autocomplete.

PR #90, opened 2023-04-09
Add share button

### LINKED ISSUE #83: [Feature Request] Data sharing functionality
# Overview

Describe the motivation behind your desired functionality.

The URL of the website currently reflects the user's provided search terms, updating as they add or remove options. However, this is not exactly made clear to the user.

# Proposal

Describe what changes need to be implemented to resolve this issue. Be as
detailed as possible so maintainers can best determine how to implement this
new feature.

A share button should be included somewhere in the website that provides standard sharing functionality of the current URL. This should work on mobile as well as desktop, perhaps simply copying to clipboard if on desktop, but opening the standard sharing prompt on mobile.

TyHil:
## Overview

Resolve #83

Added a share button to the top nav bar. Uses the MUI share icon. If a sharing UI is available it uses that, if not it copies the URL to the clipboard. The URL used is the current URL unless there is no data on the dashboard, in which case the home page (not the dashboard page) URL is shared.

### Question to answer

Should the color of the icon be white or the current dark purple?

### DIFF
--- a/components/navigation/topMenu/topMenu.tsx
+++ b/components/navigation/topMenu/topMenu.tsx
@@ -1,19 +1,22 @@
-import { Icon } from '@mui/material';
+import { IconButton } from '@mui/material';
 import { LogoIcon } from '../../icons/LogoIcon/logoIcon';
-import HandymanIcon from '@mui/icons-material/Handyman';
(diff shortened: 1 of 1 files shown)


iamwood (approved):
Awesome! Works great.

PR #91, opened 2023-04-10
MUI Icons

TyHil:
## Overview

Reformats the `renderOption` of the search bars to use MUI InputBase components and thus allow `startAdornment`. I used the MUI search icon which looks cleaner than our existing one and thus removed `searchIcon.tsx`.

## What Changed

`renderOption` in `searchBar.tsx` and `splashPageSearchBar.tsx`. Also some styles in those files plus `expandableSearchGrid.tsx`. Removal of SearchIcon related files.


iamwood (approved):
These are good changes. Standardization and precision of our styling is always preferred, and the MUI icon looks good.

PR #96, opened 2023-04-14
Update CardContents.tsx

George-Zhang5246:
Changed difficulty scale to 1 to 5 from previous 0 to 4, since I messed up the range in the previous change.


iamwood (approved):
This change makes sense, good to go.

PR #97, opened 2023-04-15
Number prefix query order

TyHil:
## Overview

Added new query formats to the autocomplete graph that let you type course number before prefix, like 1100 ECS. Now anywhere you could type `<prefix><number>` you can type `<number>[<prefix>]`. The prefix is optional.

## What Changed

Fixed a bug where the autocomplete was only checking the first character of each node.
Switched to a priority queue for bfs.
Removed duplicate autocomplete results.
Used a lot of paths in the DAG that split and rejoin to implement new formats.

### Query formats now supported

`<prefix>[<number>| <number>[.<section>][ <professorLast>|(<professorFirst> <professorLast>)]]`
`(<number>|<number> )[<prefix>][.<section>][ <professorLast>|(<professorFirst> <professorLast>)]`
`(<professorLast>|<professorFirst> <professorLast>) ((<prefix> <number>|<prefix><number>)|(<number><prefix> |<number><prefix>))`


Jake3231:
One small thing, it looks like autocomplete highlighting does not work with the format `1200CS` where it does with `CS1200`. Is this expected?

Screenshot 2023-10-18 at 6 21 55 PM
1d771c-7d6d-411e-9063-f6f0520a47a6">

Screenshot 2023-10-18 at 6 20 18 PM

pages/api/autocomplete.ts
```diff
@@ -25,36 +25,92 @@ graph.updateEachNodeAttributes((node, attr) => {
...
+    return this.items[0];
+  }
+  isEmpty() {
+    return this.items.length === 0;
+  }
+}
+
+function bfsRecursionToNextData(queue: PriorityQueue) {
+  const queueItem = queue.dequeue();
+  //console.log(graph.getNodeAttribute(queueItem?.data?.node, 'c'));
```
Jake3231, line 73: This commented line should probably be removed.

pages/api/autocomplete.ts
```diff
@@ -65,24 +121,26 @@ function bfsRecursion(queue: QueueItem[]) {
   let returnData = false;
 
   //# of characters matched
-  const nodeCharacters = graph.getNodeAttribute(queueItem?.node, 'c');
+  const nodeCharacters = graph.getNodeAttribute(queueItem?.data?.node, 'c');
   let matches = 0;
   while (
     matches < nodeCharacters.length &&
-    queueItem?.characters?.[0] === nodeCharacters[0]
+    matches < queueItem?.data?.characters?.length
```
Jake3231, line 128: Does this work where `characters` is undefined/nil?

pages/api/autocomplete.ts
```diff
@@ -65,24 +121,26 @@ function bfsRecursion(queue: QueueItem[]) {
...
-    /*queueItem?.characters?.[0] ===
-    graph.getNodeAttribute(queueItem?.node, 'c')*/
     nodeCharacters.length === matches ||
-    queueItem?.characters?.length === matches
+    queueItem?.data?.characters?.length === matches
   ) {
     //full match or end of characters to match but all matched
-    //console.log('match: ', queueItem?.characters, queueItem?.characters?.length === 1);
-    if (queueItem?.characters?.length <= nodeCharacters.length) {
+    //console.log('match: ', queueItem?.data?.characters, queueItem?.data?.characters?.length === 1);
```
Jake3231, line 142: Commented line

pages/api/autocomplete.ts
```diff
@@ -91,35 +149,45 @@ function bfsRecursion(queue: QueueItem[]) {
...
   ) {
     //partial match
     queueToNext = true;
     returnData = true;
   }
 
   if (queueRecursion) {
-    graph.forEachOutNeighbor(queueItem?.node, (neighbor) => {
+    graph.forEachOutNeighbor(queueItem?.data?.node, (neighbor) => {
       //console.log('queue: ', graph.getNodeAttribute(neighbor, 'c'));
```
Jake3231, line 161: This should probably be removed also

pages/api/autocomplete.ts
```diff
@@ -25,36 +25,92 @@ graph.updateEachNodeAttributes((node, attr) => {
...
     });
   }
   return;
 }
 
-function bfsRecursion(queue: QueueItem[]) {
-  const queueItem = queue.shift();
-  if (typeof queueItem === 'undefined') {
+function bfsRecursion(queue: PriorityQueue) {
+  const queueItem = queue.dequeue();
```
Jake3231, line 112: In order to remove all of the `queueItem`s, could a type guard be implemented at the top level of this function and provide another variable for use within the function?

scripts/generateAutocompleteGraph.js
```diff
@@ -0,0 +1,388 @@
...
+      );
+      if (typeof preExisting === 'string') {
+        if (characters.length <= 1) {
+          //console.log('found: ', characters[0], 'end');
+          if (typeof data !== 'undefined') {
+            graph.setNodeAttribute(preExisting, 'd', data);
+          }
+          return preExisting;
+        }
+        //console.log('found: ', characters[0]);
```
Jake3231, line 131: I'd consider removing commented lines used for debugging.

scripts/generateAutocompleteGraph.ts
```diff
@@ -191,51 +194,67 @@ fetch('https://catfact.ninja/fact', { method: 'GET' })
       return addSearchQueryCharacter(newNode, characters.slice(1), data);
     }
 
+    function addWithParents(
```
Jake3231, line 197: It would probably be worth adding a short (even few words) comment for what this method does.

Jake3231 (commented):
Overall looks good! I left a few comments on small things. One general thing is that I'd add comments above functions with more verbose or non-obvious signatures which explain what they do.

TyHil:
`1200CS` highlighting fixed

pages/api/autocomplete.ts
TyHil: I don't think `characters` will ever be undefined. I can add a check (`typeof queueItem?.data?.characters?.length !== 'undefined'`) that will discard that search result if it ever is. Do you think that's a good idea?

scripts/generateAutocompleteGraph.ts
TyHil: Great idea, added in latest commit

pages/api/autocomplete.ts
Jake3231: My mistake, I missed a ? character! Instead of the optional `queueItems` could this be effectively unwrapped at a high level for use below?

TyHil: Like instead of passing in `queue`, passing in `queue.dequeue()` and `queue.enqueue`?

pages/api/autocomplete.ts
Jake3231: Yeah, it makes sense that it shouldn't be undefined, though just wanted to be sure that couldn't happen. If we know that `data` can't be undefined then this is probably fine as written.

pages/api/autocomplete.ts
Jake3231: I'm not sure the best way to do this in TypeScript, the way I might write something like this in Swift is
```if let queueItem = queue.dequeue() {
          queueItem is always non-nil here
          } else {
          handle the nil case
          }
 ```
 
 The idea would be to guarantee that queueItem is never nil, so we can use `queueItem` instead of the optional `queueItem?`. Not a huge deal, and I think it's fine as it, but could be something to consider if this type of thing is doable in TypeScript.

TyHil: Good idea, that let me remove all the `?.`s. Done

Jake3231 (approved):
Thanks for making those changes, looks good to go!

PR #98, opened 2023-04-16
Blank search updates

TyHil:
## Overview

Removed suggestions on the homepage before typing anything.
Set suggestions on homepage to be based on last inputted query. So if you add "CS 1200" to the expandable search grid and click on the search bar, the immediate suggestions that pop up will be things like "CS 1200 John Cole", "CS 1200 Prof name", and so on. This is like a primitive related classes box.

## What Changed

Made other bug fixes along the way and improvements based on the MUI documentation but primary changes in are in the search bar files.


TyHil:
Integrated Common Types and Functions (PR #103) into branch.

Jake3231:
The new behavior seems improved! @TyHil Just to verify, is all new behavior working as expected with the latest changed to develop?

TyHil:
@Jake3231 Yup, all working

Jake3231:
@TyHil Is it correct that this has no conflicts with the changes in #97 ?

Jake3231 (approved):

TyHil:
@Jake3231 I don't know. They both work independently and I don't see any reason they shouldn't work together. Can we merge one into dev so I can run some tests and see?

Jake3231:
Sounds good, it looked like that to me as well. Once one is merged we can verify.

iamwood (approved):
Looks good to merge

PR #99, opened 2023-04-17
Search bolding

TyHil:
## Overview

Add bolding to searches. If you search "CS 12" for example, results will look like **CS 12**00, **CS 12**00 John Cole, etc.
This uses, as suggested by MUI, the autosuggest-highlight library: https://github.com/moroshko/autosuggest-highlight.

## What Changed

Changed `renderOption` in both search bars. Removing the `className` property on it also fixed the highlighting issues.
I used regex to split any CS1200 or 1200CS into CS 1200 or 1200 CS respectively. Also supporting partial numbers like CS1, CS12, and CS120 and numbers with letters in them like CS5V81 of which I think there is only ever the possibility of a V in the second position.


Jake3231 (approved):
Looks good! Just needs conflict resolution and should be good to go

PR #100, opened 2023-04-19
Remove box+whisker+stdev

TyHil:
## Overview

Removes all graphs but the standard grades bar chart.
Moved average GPA to the expandableSearchGrid.

## What Changed

Pass data to expandableSearchGrid and then searchTermCard and render properly. Add disclaimer tooltip on how dropped grades are excluded from average.


iamwood (approved):
Looks good to merge.

PR #103, opened 2023-04-25
Common Types and Functions

TyHil:
## Overview

Create common type and function files for SearchQuery, searchQueryLabel, and searchQueryEqual. Add appropriate imports into pages, components, and APIs.

## What Changed

New folders and files for SearchQuery, searchQueryLabel, and searchQueryEqual and had to move the GraphProps file into its own folder.


iamwood (approved):
Good change, reducing redundant type declarations is always a good idea.

PR #105, opened 2023-04-26
Feature/hide data icon button

TyHil:
## Overview

Allow hiding queries from all graphs and professor tab

## What Changed

Added a button in each `searchTermCard` to toggle visibility. This generates a visibility array in the `expandableSearchGrid` which is then passed to the dashboard. The dashboard does not rerun data fetching and instead excludes datasets at the last time possible, before becoming props for the graphs/prof cards.


Jake3231:
Overall looks good, and seems very useful! From a design perspective, is the visibility toggle discoverable?

TyHil:
I just tabbed through and it could access the toggles

Jake3231:
The biggest thing I noticed is that it seems to only appear when you hover over the color of a selected course, and there's no indication that the toggle is hidden there.

TyHil:
Good point. I remember thinking it looked weird to always have the Visibility icon there but its better to be clear. New commit should fix that

Jake3231 (approved):
Overall looks good! cc @iamwood as well on design

iamwood (approved):
Looks good to merge

PR #126, opened 2023-11-09
Bump @babel/traverse from 7.21.4 to 7.23.2 (#120)

AdiC03:
Updated the professor card (All the data syncs up still; nothing is hard coded except for course review). 
Note: We still need to add the percentage of people who completed a course review.


Jake3231 (changes requested):
Looks like this change is just updating dependencies, and does not add the new component. Are there missing files?

Jake3231:
See #137

PR #127, opened 2023-11-13
Redesign homepage

TyHil:
## Overview

List the GitHub issues containing the issues relevant to this pull request.

[For example, "Resolves #1, closes #2, fixes #3", etc.]

Describe your changes here at a high level, describing how this PR fits into the
rest of the project.

## What Changed

Go more into detail about key files that were modified and why they were
updated. Do not list every file that was modified; only note the ones most
relevant to this feature or bug fix.

## Other Notes

If were roadblocks encountered during development that remain unresolved or any
future additions or changes to make, note them here. Otherwise. Feel free to
delete this section if it isn't needed.

PR #124, opened 2023-11-02
Update fonts

TyHil:
Update fonts from defaults Inter and Kallisto per the style guide

PR #123, opened 2023-11-01
Standardize favicon and open graph data across Nebula

TyHil:
I was updating the Trends icons and decided to standardize the favicons and open graph data on all the Nebula sites. This PR updates the favicons to the new Nebula logo and adds/standardizes the [Open Graph](https://ogp.me/) metadata.

Every site should now follow the same template for the head tags in `_app.tsx` and `_document.tsx` which all link to the same corresponding icons. I also stored this icons in [this drive link](https://drive.google.com/drive/folders/1zHHQVsNNeHIr48K0DzpI6KZpIDqYFLgz?usp=sharing).

Other sites PRs:
https://github.com/UTDNebula/utd-trends/pull/123
https://github.com/UTDNebula/website-v2/pull/19
https://github.com/UTDNebula/jupiter/pull/175
https://github.com/UTDNebula/planner/pull/731

PR #128, opened 2023-11-13
Update colors

TyHil:
Update colors to new design specs. Basically copied this from the new website.

Note: This breaks a lot of colors! But everything it breaks needs to be redesigned anyways

### DIFF
--- a/pages/_app.tsx
+++ b/pages/_app.tsx
@@ -22,7 +22,7 @@ function MyApp({ Component, pageProps }: AppProps) {
     palette: {
       mode: prefersDarkMode ? 'dark' : 'light',
       primary: {
-        main: '#7486ce',
+        main: '#573dff',
       },
     },
   });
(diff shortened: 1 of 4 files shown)


Jake3231:
To clarify, this is for the redesign, not the current design?

TyHil:
Yes it is. Should I replace this, #123, and #124 with PRs to merge into `redesign`?

Jake3231:
I would switch to merge into `redesign`, which you can do on this PR.

TyHil:
Sounds good. Would the best way to forward the changes be to make duplicate PRs targeting `redesign`?

Jake3231:
> Sounds good. Would the best way to forward the changes be to make duplicate PRs targeting `redesign`?

I would look into cherry picking the change into redesign.

Jake3231 (approved):
Approving the style changes - the biggest thing I'm a little concerned with is integration with #127 , but we can test after one is merged.

TyHil:
I branched #127 off of this branch so I could use the new colors so there shouldn't be any problems.

Jake3231:
@TyHil Is this good to merge?

PR #129, opened 2023-11-13
Remove clear on blur

TyHil:
Saw this handy prop in the MUI API and always thought it was annoying that the search cleared when you deselect it.

https://mui.com/material-ui/api/autocomplete/#Autocomplete-prop-clearOnBlur


Jake3231:
Is the expected behavior change that when you click away from the search bar without selecting something it will not remove the text field value?

Jake3231 (approved):

TyHil:
Yes. This prevents you from losing your incomplete search when navigating to another tab or doing anything else

PR #131, opened 2023-11-14
Null grade data handling

TyHil:
Instead of throwing an error when API returns `grade_distribution: null` this puts "No grade data available" in the expandable search grid makes the grade distribution `[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]` so the graph leaves a gap where it would be.

Here is an example: https://utd-trends-4tiftwexw-utdnebula.vercel.app/dashboard?searchTerms=CS+1200%2CCS+3345.HON+Ivor+Page


Jake3231:
Screenshot 2023-11-15 at 6 19 44 PM

In the case where only the course with no grade data is selected, does it make sense to show just the empty chart?

TyHil:
I'm not sure. Would the alternative be to remove the chart?

Jake3231:
That is one alternative. I’d do *something* to indicate that data is not available so that it doesn’t look broken. It may be worth asking the design channel for input.

Besides removing the chart, you could show a message on top of it saying something like “Grade data unavailable for selected courses”

TyHil:
Got it. I went with a message on top of the graph.
I went ahead and cleaned up the graph components we have since we have no plans to use anything beyond a bar graph. We can always get them back from GitHub anyways.

PR #134, opened 2023-11-21
Split professor name into first and last

TyHil:
Part of some changes with API to combat null data being returned in Trends. Split the `professorName` field into `profFirst` and `profLast` to better make API calls with.


Jake3231:
Seems good overall! What's an example of a query that didn't work before that is fixed by this PR?

TyHil:
Good question. This fixes calls for professors like Chung Hwan Kim who was previously being stored only as Chung Hwan.

pages/api/grades.ts
```diff
@@ -47,14 +49,13 @@ export default function handler(
...
+    typeof req.query.profFirst === 'string' &&
+    'profLast' in req.query &&
+    typeof req.query.profLast === 'string'
   ) {
-    url.searchParams.append(
-      'first_name',
-      req.query.professorName.split(' ')[0],
-    );
-    url.searchParams.append('last_name', req.query.professorName.split(' ')[1]);
+    url.searchParams.append('first_name', req.query.profFirst);
```
Jake3231, line 57: The API supports this also?

Jake3231 (approved):
Looks good overall

pages/api/grades.ts
TyHil: Yes it does

PR #135, opened 2023-11-21
Add professor API route for skedge

TyHil:
This will allow skedge to make all its calls to the Nebula API through Trends with its hidden API key

### DIFF
--- a/pages/api/professor.ts
+++ b/pages/api/professor.ts
@@ -1,0 +1,61 @@
+/*
+This API route is currently just for skedge
+*/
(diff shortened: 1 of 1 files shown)


Jake3231:
Is there a corresponding change in Skedge for this, or just preparing for future changes?

TyHil:
Just preparing for future changes

Jake3231 (approved):
Seems good to get ready for future changes - I assume this will need to be tweaked as Skedge develops.

PR #136, opened 2023-11-29
Feedback popup

TyHil:
This is the "frontend" for a feedback popup. I'll need to set up a backend maybe with Qualtrics.

I started by merging in the update colors branch for the redesign so colors outside the popup will be messed up.

The popup currently appears after 1 second for testing and always pops up again on reload. Before merging into redesign I'll change this to likely 60 seconds and have it remember if someone has already submitted.


Jake3231:
> Before merging into redesign I'll change this to likely 60 seconds and have it remember if someone has already submitted or closed the popup.

Does this still need to be done before merging?

Jake3231:
> Yes it does. The changes need to be made where I've left triple slashes (`///`), should I do that now?

Up to you and @iamwood , though I think it would be good to make that change now.

TyHil:
@iamwood if we still want this feature it's ready to merge

iamwood (approved):
I like this, at some point we'll just need to get the google form moved over to an official Nebula location (maybe Confluence has an equivalent?). For now this is good though

TyHil:
I could look into a Confluence API of some sort. The google sheet is owned by my nebula email if that helps.

PR #137, opened 2023-11-30
Updated Professor Card

Jake3231:
## Overview

This PR implements the updated professor card design.

(@AdiC03 new PR to change base and compare branches)

See #126

### DIFF
--- a/components/common/ProfessorCard/ProfessorCard.tsx
+++ b/components/common/ProfessorCard/ProfessorCard.tsx
@@ -42,7 +42,7 @@ export const ProfessorCard = (props: ProfessorCardProps) => {
       {/* Number of ratings */}
       <Box className="flex flex-col items-center justify-center mb-2">
         <Link
-          href={`https://www.ratemyprofessors.com/search/professors/?q=${encodeURIComponent(
+          href={`https://www.ratemyprofessors.com/search/professors/1273?q=${encodeURIComponent(
             props.name,
           )}`}
           target="_blank"
--- a/components/common/ProfessorCard/ProfessorCard.tsx
+++ b/components/common/ProfessorCard/ProfessorCard.tsx
@@ -1,3 +1,7 @@
+// Only problem is there's a black sillhoute that appears
+// Because of the smaller professor card. Most likely due to the css element.
+// Need api key
+
 import { Box, Typography } from '@mui/material';
 import Link from 'next/link';
 import React from 'react';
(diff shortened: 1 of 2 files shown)


components/common/ProfessorCard/ProfessorCard.tsx
```diff
@@ -9,90 +13,100 @@ type ProfessorCardProps = {
   averageDifficulty: number;
   takingAgain: number;
   numRatings: number;
+  //numEvaluation: number; (Need to include percentage for evalution)
```
Jake3231, line 16: @AdiC03 Is this expected?

components/common/ProfessorCard/ProfessorCard.tsx
```diff
@@ -1,3 +1,7 @@
+// Only problem is there's a black sillhoute that appears
+// Because of the smaller professor card. Most likely due to the css element.
```
Jake3231, line 2: This should be resolved before merging if possible.

Jake3231 (commented):
Left a few questions @AdiC03

Jake3231:
@AdiC03 Do you expect to be able to make changes to this PR ahead of merging?

AdiC03:
> @AdiC03 Do you expect to be able to make changes to this PR ahead of merging?

Not really. There is a small problem regarding the background css but I can't really seem to figure it out (very minor issue)

Jake3231:
> Not really. There is a small problem regarding the background css but I can't really seem to figure it out (very minor issue)

Have you asked around in #trends-engineering about addressing that?

Jake3231:
@TyHil What do you suggest for this PR?

TyHil:
Looks like the "Based on x ratings" link opens to a search for the professor in any school, not just UTD.
Also the "Would Take Again" box shows as a rectangle for me, we should fix that to a square like in the Figma.

Other than that, this looks phenomenal!

components/common/ProfessorCard/ProfessorCard.tsx
TyHil: I think this ought to be fixed when we put all the components together as this isn't really a problem with the professor card but just a side effect of displaying it in the current UI.

TyHil (changes requested):
Looks like the "Based on x ratings" link opens to a search for the professor in any school, not just UTD.
Also the "Would Take Again" box shows as a rectangle for me, we should fix that to a square like in the Figma.

Other than that, this looks phenomenal!

Jake3231:
Is this still relevant @iamwood ?

PR #132, opened 2023-11-14
Late 2023 Production update

Jake3231:
- Bumps [@babel/traverse](https://github.com/babel/babel/tree/HEAD/packages/babel-traverse) from 7.21.4 to 7.23.2.
- Updates Nebula logos
- Updates fonts across the app
- Provides an activity indicator in the search autocomplete
- No longer clear search field input when moving focus
- Improve handling when no grade data is available for a course (#131, #134)
- Project cleanup (#133)
- Improve appearance of some text in the dark appearance (#140)

PR #139, opened 2023-11-30
Late 2023 Production update -> redesign

TyHil:
Ditto of PR #132 just to update redesign with recent changes.

I believe this should be a normal merge, not a squash and merge


TyHil:
Yeah I'm not sure how tho, since we can't make a merge commit on develop. What do you think?

Jake3231:
Since we're targeting redesign, shouldn't we be able to do normal conflict resolution locally?

TyHil:
Oh yeah I guess so, no PR necessary. I can do that.

TyHil:
Done

PR #141, opened 2023-12-10
Redesign layout

TyHil:
Strips a lot of functionality from the website so it can be readded in different places once components are finished.


Jake3231:
@TyHil What's the latest status on this PR?

TyHil:
Waiting on components to be finished and merged into redesign first

PR #142, opened 2024-01-30
Slight autocomplete performance improvement

TyHil:
Slight improvement to autocomplete response time by reducing edge key size.
Also reduces the size of `data/autocomplete_graph.json` by about 1.5 Mb.

The main choke point, which is not addressed in this PR, is with `graph.import(autocompleteGraph as object);` in `pages\api\autocomplete.ts`. This line runs when the API is first accessed and when it hasn't been used in a while and Vercel has shut down the route. It takes approximately 500ms on my machine.


iamwood (approved):
This change makes sense, and will probably improve performance to a slight degree.

PR #143, opened 2024-02-06
Autocomplete performance improvement

TyHil:
The main slowness with the autocomplete system was waiting for Vercel to "spin up" the endpoint when it hadn't been used in a while. This change makes a call to the endpoint immediately on page load so it's running by the time you get to the search bar, hopefully.

### DIFF
--- a/components/common/SearchBar/searchBar.tsx
+++ b/components/common/SearchBar/searchBar.tsx
@@ -74,6 +74,10 @@ export const SearchBar = (props: SearchProps) => {
     }
   }, [open, inputValue]);
 
+  useEffect(() => {
+    fetch('/api/autocomplete');
+  }, []);
+
   return (
     <>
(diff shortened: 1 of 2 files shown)


Jake3231:
To be sure I've got it, this simply accesses the endpoint upon page load to get started early? Do we just waste resources if the user never clicks the search bar?

TyHil:
Correct. We do waste resources if the user never clicks the search bar. I was assuming, and correct me if I'm wrong, that our Vercel hosting is free so it doesn't matter all that much.

Jake3231 (approved):

iamwood (approved):
I agree with @Jake3231 that we can add this now, but we do need to find a better way to maintain the endpoint in a low-cost manner.

Jake3231:
See #145 tracking this

Jake3231:
@TyHil @iamwood We should be good to merge, though with that follow up issue tracking future changes.

PR #152, opened 2024-03-29
Add tag functionality and filters to search bar

### LINKED ISSUE #149: [Feature Request] Dropdowns for searching
# Overview

This part of the main navigation bar should provide users with a clear way of selecting options from a multi-tiered set of options. Our dataset(s) can be split and or grouped according to multiple intersecting categories (most notably professors and courses), so ideally these menus would be dynamically rendered according to an easily modified tiered registry.

# Proposal

The final design of this has not been established, so a working prototype of this sort of multi-tiered option selection should be a good starting place to generate ideas from.

## Supplementary Information

This feature has not been entirely built-out yet, so if you are interested in working on it, please reach out in the Discord and or make a comment here in this Issue.

### LINKED ISSUE #153: [Feature Request] Chip-based search bar with filters
# Overview

The search bar's functionality has been reoriented to be more flexible and accommodate the addition of numerical filters. The user will now add identifiers (course prefix and number, professor name, or course name) as chips, suggested when they begin typing, and (optionally) adjust range filters for numerical values (average GPA, RMP rating, RMP difficulty). When the user initiates the search (hits enter, clicks spyglass), it will take them to the dashboard with the results in the main list. These results will always be a course and a professor together. If the user entered just a course, then it will show that same course with all the professors who have taught it. If the user entered just a professor, then it will show that professor with all the courses they have taught.

# Proposal

The current autocomplete search bar utilizes the React MUI Autocomplete component, with some additional elements added to facilitate the unique radix tree solution for how searching is performed currently. These parts can be removed to instead suggest the individual identifiers when the user is typing. The filters should be added below as in the `V1.1 Homepage` Figma page. One option for presenting this filter selection is with a MUI range slider, documented here: https://mui.com/material-ui/react-slider/#range-slider.

# Supplementary Information

Be sure to check out the `V1.1 Homepage` page in Figma to get an idea of the look of the new search bar: https://www.figma.com/file/U6d2F910K3VwnLqRc7ifEJ/Trends-Wireframing?type=design&node-id=52-598&mode=design.

TyHil:
Resolves #153 
Resolves #149 

Tags are created when you hit enter or if you hit space if there's only one autocomplete option left. This works in 2 parts.
- If the suggestions are already loaded and you hit space, the tag is added.
- If a suggestion loads late, notices the last character was a space it adds the tag and keeps the remaining text.
Had to fix a lot of bugs in the autocomplete graph generation code that were making prefixes chain together.

Filters are implemented as dropdowns in a separate component.


iamwood (approved):
Looks good to merge.

PR #156, opened 2024-05-05
Redesign navbar

TyHil:
Brings topMenu.tsx up to date with the current design.


iamwood (approved):
Looks good, might want to change the Min and Max filters to be some sort of range selector in the future, but will definitely work as is for now

PR #157, opened 2024-05-05
Colorful graph bars

TyHil:
Since we'll be having a lot more graphs with only one series, pretty much all of them except the compare graph, I thought it would be nice for them to have some color.

This PR makes any graph showing only one data set use the rainbow colors from skedge


iamwood (approved):
Seems like a nice improvement, LGTM.

PR #158, opened 2024-05-05
Move fetch with cache

TyHil:
Moves the logic to make and cache a fetch request to outside `dashboard/index.tsx`. This just reduces some complexity in that file as we rewrite it and makes the caching available in other files if desired.


iamwood (approved):
LGTM

PR #164, opened 2024-08-25
Remove nodag stuff, causes confusion

TyHil:
We don't need to be building the autocomplete graph on every deployment rn so I removed the package scripts that do that.

Once there's an API endpoint to get the data needed for the autocomplete graph we can set vercel to do `npm run buildautocomplete && npm run build` on deployment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,12 +2,9 @@
   "name": "UTD-Trends",
   "private": true,
   "scripts": {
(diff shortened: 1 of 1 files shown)


AbhiramTadepalli (approved):
_"it makes it so you don't have to add :nodag to the end of stuff to avoid rebuilding the autocomplete graph"_

PR #167, opened 2024-08-26
Move code to src

TyHil:
This moves all website related code to an `src` folder. Doing so separates this code from config files and such, making them easier to distinguish and allowing you to know exactly what is being deployed.

This change also bring Trends in line with Jupiter and Planner, which do the same.


AbhiramTadepalli (approved):

AbhiramTadepalli:
@TyHil can you resolve conflicts?

PR #168, opened 2024-08-26
Use Skedge's RMP scraper

### LINKED ISSUE #151: [Feature Request] Use skedge RMP fetch code
# Overview

Skedge has a better RMP scraper that is able to pull more information such as professor tags. It does this by searching for the professor in the same way that we do but then goes a step further. Instead of scraping the limited information directly from this page Skedge uses the RMP graphql endpoint to pull more data based on the found professor.

# Proposal

We should be using this as well to display the same information on Trends.

## Supplementary Information

The Skedge RMP scraper can be found here: https://github.com/UTDNebula/skedge/blob/develop/src/data/fetchFromRmp.ts

### LINKED ISSUE #178: [Question]: Choose RMP for a professor with 2 profiles
@TyHil how do you choose the correct RMP for a professor with 2 entries, like:
Dr. Octavious Smiley: https://www.ratemyprofessors.com/professor/2941177 & https://www.ratemyprofessors.com/professor/2982353
Dr. Scott Dollinger: https://www.ratemyprofessors.com/professor/2523207 & https://www.ratemyprofessors.com/professor/2651411

### Code of Conduct

TyHil:
Resolves #151 
Resolves #178 

Uses a modified version of Skedge's RMP scraper to access the graphql endpoint and get more RMP data.

Data is not implemented into the dashboard, only the existing functionality is preserved.

Instead of getting the professor's ID from scraping RMP search page with regex, this now only uses one graphql call! A list of professors is returned and only those at UTD and with matching name are considered. The one with the most ratings is then chosen.


AbhiramTadepalli (approved):
Well done Tyler!

PR #169, opened 2024-08-27
Re add overall in search results

TyHil:
As discussed in last Trends meeting, the move to the combos api route removed the overall results for single prof or course searches. This PR readds that by just adding the searched value locally to the results of the call to the combo api.


pages/dashboard/index.tsx
AbhiramTadepalli, line 1: @TyHil So this is just:
Instead of mapping through the searchTerms and fetchWithCache, it is calling combosSearchResultsFetch for each searchTerm?

AbhiramTadepalli (approved):

PR #190, opened 2024-09-17
Fix name/credit hours for special courses

### LINKED ISSUE #173: [Bug]: Course Description includes the course's name when it contains special characters.

The course description should start with the actual description for every course.

### What happened?

For courses with special characters in their name (parenthesis, slashes) the description includes the course name again which is repetitive and not consistent with other courses.

### Version

Beta 0.1.0

_No response_

### Relevant log output

_No response_

### Code of Conduct

### Tasks
- [ ] Recognize that a course name without special characters is equal to the name at the start of the description
- [ ] Filter out the course name at the start of the course description
- [ ] Fix the course name (add in special characters)

### LINKED ISSUE #183: [Bug]: Course name for courses like CS 4V95
### What happened?

CS 4V95 is a "special topics" course: "Undergraduate Topics in Computer Science Software Engineering". Right now, it shows _"Undergraduate Topics in Computer Science - Network Traffic Generation"_ as the course's name; this is one instance of the class, but doesn't represent the entire class.

### What should happen
Any course that has a variable (special topic) should have a general course name, not the specific instance of the course that comes after the dash _in this case_

- [x] Identify all the "special topics" courses like (variable number of credit hours, like CS 2V95)
- [x] Find a way to remove the specific info from the course name
- [ ] Verify it works for all courses like this

### Version

Beta 0.1.0

_No response_

### Relevant log output

_No response_

### Code of Conduct

egsch:
## Overview
Resolves #183, also resolves #173.

## What Changed

In CourseOverview.tsx, I replaced the course name we get from the Nebula API with one parsed from the course description. I did the same with credit hours **only when the course has variable credit hours.** I replaced our parsing technique to use some simple regex that accounts for a variable number of credit hours. 

## Other Notes

We still might need some further testing to make sure none of our other course names/descriptions were affected by this. The special courses listed in the issue all have the correct name now, and other random courses seem to be working correctly.

### DIFF
--- a/src/components/common/CourseOverview/courseOverview.tsx
+++ b/src/components/common/CourseOverview/courseOverview.tsx
@@ -28,10 +28,16 @@ function parseDescription(course: CourseData): {
   creditHours: string;
 } {
   //extracts info from the course description and formats it
-  const [descriptionBeforeCreditHours, descriptionAfterCreditHours] = course.description.split(/\([\d-]{0,3} semester credit hour[s]?\)/);
+  const [descriptionBeforeCreditHours, descriptionAfterCreditHours] =
+    course.description.split(/\([\d-]{0,3} semester credit hour[s]?\)/);
   let formattedDescription = descriptionAfterCreditHours || course.description;
-  const courseTitle = descriptionAfterCreditHours ? descriptionBeforeCreditHours.split(course.course_number + " - ")[1] : course.title;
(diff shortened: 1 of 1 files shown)


AbhiramTadepalli (approved):
Wow so good!

AbhiramTadepalli:
Great job @egsch

Issue #170, opened 2024-08-29
[Bug]: Crash when sorting RMP


AbhiramTadepalli:
### What did you expect to happen?

Properly sort without crashes even when all data is not loaded

### What happened?

Sort by gpa twice, then sort by rmp → crash (some uncaught undefined’s)

Am not able to replicate currently, might have something to do with not-yet-loaded rmp data.

### Version

Beta 0.1.0

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

AbhiramTadepalli:
New log from develop branch:

> Grades data for CS 6360 Aditya Srivastava Error: null data
    G NextJS
[dashboard-f8c958d8181962c9.js:1:33684](https://dev.trends.utdnebula.com/_next/static/chunks/pages/dashboard-f8c958d8181962c9.js)
Empty string passed to getElementById(). 3 [153-5bbd237ece401776.js:1:23350](https://dev.trends.utdnebula.com/_next/static/chunks/153-5bbd237ece401776.js)
Grades data for CS 6363 Partha De Error: null data
    G NextJS
[dashboard-f8c958d8181962c9.js:1:33684](https://dev.trends.utdnebula.com/_next/static/chunks/pages/dashboard-f8c958d8181962c9.js)
Empty string passed to getElementById(). [153-5bbd237ece401776.js:1:23350](https://dev.trends.utdnebula.com/_next/static/chunks/153-5bbd237ece401776.js)
Grades data for CS 6390 Ravi Prakash Error: null data
    G NextJS
[dashboard-f8c958d8181962c9.js:1:33684](https://dev.trends.utdnebula.com/_next/static/chunks/pages/dashboard-f8c958d8181962c9.js)
Empty string passed to getElementById(). 4 [153-5bbd237ece401776.js:1:23350](https://dev.trends.utdnebula.com/_next/static/chunks/153-5bbd237ece401776.js)
Grades data for CS 5349 Error: null data
    G NextJS
[dashboard-f8c958d8181962c9.js:1:33684](https://dev.trends.utdnebula.com/_next/static/chunks/pages/dashboard-f8c958d8181962c9.js)
Grades data for CS 5349 Error: null data
    G NextJS
[dashboard-f8c958d8181962c9.js:1:33684](https://dev.trends.utdnebula.com/_next/static/chunks/pages/dashboard-f8c958d8181962c9.js)
Grades data for CS 5349 Neeraj Gupta Error: null data
    G NextJS
[...]

PR #207, opened 2024-09-27
Fixed sorting bug

RVC13850:
## Overview
Fixed issue #170 where it would crash when sorted more than 2 times.

## What Changed
Added a check to make sure aGrades and bGrades objects were defined.


TyHil:
This seems to have removed the check if they're both undefined/not loaded. It was this:
```
if (aGrades.state !== 'done' && bGrades.state !== 'done') {
  return 0;
}
```
If there's not a reason, can you bring it back (for grades and rmp) and apply your changes from the other if statements to these as well.

RVC13850:
There was no particular reason. I added it back and its working so no issues

TyHil:
Looks excellent then! I'll let @AbhiramTadepalli review

AbhiramTadepalli (approved):
Looks great!

PR #222, opened 2024-10-02
New compare table

### LINKED ISSUE #209: [Feature Request] New Compare component
# Overview

Redesign the compare table to look like the new [Figma design](https://www.figma.com/design/OmGLQNdZBrgWGfamGjkci5/Redesign-Mockups?node-id=0-1&node-type=canvas&t=jfQQFef771aVynF4-0):
image

# Proposal

Steps:
- [x] Design Compare card component (vertical)
    - [x] Try to implement the color gradient in the card as shown above
    - [x] Use static color ranges instead of dynamic colors for each piece of data (base the colors off of what is on the search result graph)
- [x] Redesign Compare Table component
- [x] Ensure all the data is formatted correctly (with skeletons on load)
- [x] Sorting (horizontal means left-most is the first result/top)

## Supplementary Information

Needs to be done by launch.

TyHil:
## Overview

Resolves #209 

This turned out to be quite a bit more difficult than expected. Why? Because you can't build a table by columns, only by rows. So you can't make a component for each query added to compare because the line wraps in each ones title get everything out of line. A table manages these line wraps. So this implements a component for each row. As in there's a component made to show the GPA info, the rmp rating, ... 

Other than that the gradient implementation went well and the sorting works.

Also copies over that fix to the sorting bug from PR #207


AbhiramTadepalli:
To-do
- [ ] Bottom padding (the horizontal scroll bar is a few pixels above the bottom of the components in my instance
- [ ] Does there need to be a way to get back to the default sort (maybe a sort by color?)
- [ ] Make sure the arrows point the correct way by default
- [ ] Question: Should each component have the same width
@TyHil

TyHil:
> Bottom padding (the horizontal scroll bar is a few pixels above the bottom of the components in my instance

Looks fine on my machine, can you send a screenshot?

> Does there need to be a way to get back to the default sort (maybe a sort by color?)

You can click through the the sort options to deselect and it returns to default

> Make sure the arrows point the correct way by default

Done

> Question: Should each component have the same width

Done

AbhiramTadepalli (approved):
Beautiful exquisíte renditión of the compárison tâbular design.

PR #223, opened 2024-10-03
Add prof RMP link

TyHil:
Adds a link to Rate My Professors like on the old Trends and on Skedge.

No matter what we scrape from RMP, linking to the site will still always hold value to users so this is a good thing to do.


AbhiramTadepalli (approved):
Great!

PR #224, opened 2024-10-03
Fixed Issue 195 - Implement column sort UI (indicating that a column is sortable) 

### LINKED ISSUE #195: [Feature Request] Implement column sort UI (indicating that a column is sortable)
# Overview

Indicates clearly that the columns are sortable

# Proposal

Sorting already works. You just need to always show the sort arrow icon.

## Supplementary Information

The selected sort icon should be darker/"selected" whereas every other icon is greyed-out/"unselected"

GovindRangappa:
## Overview

Resolves #195 

Changes make column header sorting arrow(next to GPA and Rating column headers) visible by default instead of only when hovered over or clicked

## What Changed

Only the "searchResultsTable.tsx" file was edited. Added CSS rules that overrides the default CSS rules for TableSortLabel that makes the arrow not visible by default.

### DIFF
--- a/src/components/common/SearchResultsTable/searchResultsTable.tsx
+++ b/src/components/common/SearchResultsTable/searchResultsTable.tsx
@@ -334,7 +334,12 @@ const SearchResultsTable = ({
                   }}
                   sx={{
                     '& .MuiTableSortLabel-icon': {
-                      opacity: '1 !important', // Ensure the arrow is always visible
+                      opacity:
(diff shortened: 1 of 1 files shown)


TyHil:
@GovindRangappa, thinking the arrow should be greyed out so there's a difference when you click on it and it becomes completely white. `opacity: 0.5;` by default would be good if possible.

AbhiramTadepalli (approved):
👍

PR #226, opened 2024-10-03
Fix sort arrow color

TyHil:


AbhiramTadepalli (approved):
You have _not_ failed this city

PR #227, opened 2024-10-03
Keep RHS tabs mounted

### LINKED ISSUE #186: [Bug]: Transition from Compare tab to Professor tab has lag time
### What happened?

When 3 tabs are available, once all data has loaded. Going from Tab 3 (compare) to Tab 1 (professor) causes some lag time.

_Could_ have something to do with loading-unloading of the Tab Components in the carousel

- [ ] Ensure that an API call is not being made every time you switch tabs
- [ ] Fix the lag time

### Version

Beta 0.1.0

_No response_

### Relevant log output

_No response_

### Code of Conduct

TyHil:
Resolves #186 

This prevents a new API call for each tab switch as the component is not unmounted every time you switch away, just set to `display: none`.


AbhiramTadepalli (approved):
Simple solution and works well!

PR #230, opened 2024-10-05
Make RHS sticky when scrolling through search results

### LINKED ISSUE #212: [Feature Request] RHS should stay on the screen when scrolling through results
# Overview
If you have a long list of search results, the RHS disappears as you get to the bottom of the list.

# Proposal

It is a lot more helpful if the RHS stays visible regardless of where the user is in the list of LHS search results

## Supplementary Information

idk the right term for this, maybe it is "anchored" to the screen?

egsch:
## Overview

Resolves #212
Makes sure RHS is visible through the entire page.

## What Changed
I made the RHS sticky. I applied the sticky property to the bottom of the grid item because I figured that when the RHS is taller than the screen, users would want to be able to see the bottom of the RHS before scrolling to the bottom of the page.

This is a little bit hacky (and has the opposite problem when scrolling back up, especially on the compare view), so if y'all think it's better we can also use the traditional sticky prop:
<details>
<summary>Expand to see GIF with traditional prop</summary>
With traditional sticky prop:

scrollStickyTop
</details>

<details>
<summary>GIFs with my version for comparison</summary>
With custom sticky prop:

scrollStickyBottomShort

scrollStickyBottom
</details>

This doesn't seem to affect mobile from my testing.

Also I know CSS can be weird--I just tested this on Chrome and Firefox, so if someone wants to try it with Safari that would be good!

### DIFF
--- a/src/pages/dashboard/index.tsx
+++ b/src/pages/dashboard/index.tsx
@@ -740,16 +740,12 @@ export const Dashboard: NextPage = () => {
               removeFromCompare={removeFromCompare}
             />
           </Grid>
-          <Grid
-            item
-            xs={false}
-            sm={5}
-            md={5}
-            className="sticky bottom-2 h-min gridsm:min-h-screen w-full"
-          >
-            <Card>
-              <Carousel names={names}>{tabs}</Carousel>
-            </Card>
+          <Grid item xs={false} sm={5} md={5} className="w-full">
+            <div className="sticky top-0 gridsm:max-h-screen overflow-y-auto">
+              <Card>
+                <Carousel names={names}>{tabs}</Carousel>
+              </Card>
+            </div>
           </Grid>
         </Grid>
       </>
--- a/src/pages/dashboard/index.tsx
+++ b/src/pages/dashboard/index.tsx
@@ -740,7 +740,13 @@ export const Dashboard: NextPage = () => {
               removeFromCompare={removeFromCompare}
             />
           </Grid>
-          <Grid item xs={false} sm={5} md={5} className="w-full">
+          <Grid
+            item
+            xs={false}
+            sm={5}
+            md={5}
+            className="sticky bottom-2 h-min gridsm:min-h-screen w-full"
+          >
             <Card>
               <Carousel names={names}>{tabs}</Carousel>
             </Card>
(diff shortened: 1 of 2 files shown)


AbhiramTadepalli:
> I made the RHS sticky. I applied the sticky property to the bottom of the grid item because I figured that when the RHS is taller than the screen, users would want to be able to see the bottom of the RHS before scrolling to the bottom of the page.

Totally agree.


> This is a little bit hacky (and has the opposite problem when scrolling back up, especially on the compare view)

This is the only thing I'm wary about. I'll take it over no sticky at all, but if we could find a way for RHS to fully scroll up as well that would be great.


Any insights @TyHil ?

TyHil:
What if we use the traditional sticky prop and put the RHS in a vertically scrolling div? Then it would behave like the first gif but if you want to see the rest of compare you can just scroll the RHS instead of having to scroll down to the bottom of the page. And the RHS should be able to stay at whatever scroll spot you've left it in while you scroll through the page.

egsch:
> What if we use the traditional sticky prop and put the RHS in a vertically scrolling div? Then it would behave like the first gif but if you want to see the rest of compare you can just scroll the RHS instead of having to scroll down to the bottom of the page. And the RHS should be able to stay at whatever scroll spot you've left it in while you scroll through the page.

That seems good. Just committed a version that does this.

TyHil:
beautiful

TyHil (approved):

AbhiramTadepalli:
> What if we use the traditional sticky prop and put the RHS in a vertically scrolling div? Then it would behave like the first gif but if you want to see the rest of compare you can just scroll the RHS instead of having to scroll down to the bottom of the page. And the RHS should be able to stay at whatever scroll spot you've left it in while you scroll through the page.

2 questions:
1) how does this affect mobile?
2) For a screen size/zoom like this, you'll have 2 scrollbars. I know it's intended, but is this fine in your opinion? Maybe it's something we can address further down the road, because this functionality is awesome @egsch!

egsch:
>   how does this affect mobile?

The max height/enabling of scroll only applies on larger screen sizes. And sticky does not apply either although I'm not 100% sure why. So mobile did not change at all.

>    For a screen size/zoom like this, you'll have 2 scrollbars. I know it's intended, but is this fine in your opinion? Maybe it's something we can address further down the road, because this functionality is awesome @egsch!

I think it's fine, I think a proper scrollable sticky sidebar would be better but those are pretty complicated to implement. Maybe we can check out options for those and try to add one later.

Issue #215, opened 2024-10-01
[Feature Request] Disable the search bar, when there is not valid class or professor


AbhiramTadepalli:
# Overview

Only allow clicking on "Search" when the searchTerms are not empty in both the homescreen and the dashboard

And maybe add a hover message?

PR #231, opened 2024-10-06
Guide user and prevent empty searches

AbhiramTadepalli:
## Overview

#215

Grey out search button text and show tooltip if user tries to empty search

## What Changed

Search button text is not white by default, and becomes white once the user selects at least 1 valid search term from autocomplete
Removed the alert from index.tsx that used to display above the search bar when the user tried to empty search
Transferred the text into a tooltip that pops up above the search button if the user clicks enter or the button without any search terms

### DIFF
--- a/src/components/common/SearchBar/searchBar.tsx
+++ b/src/components/common/SearchBar/searchBar.tsx
@@ -16,7 +16,6 @@ import searchQueryLabel from '../../../modules/searchQueryLabel/searchQueryLabel
 interface SearchProps {
   manageQuery?: 'onSelect' | 'onChange';
   onSelect?: (value: SearchQuery[]) => void;
-  onChange?: (value: SearchQuery[]) => void;
   className?: string;
   input_className?: string;
 }
@@ -31,7 +30,6 @@ let wasEmpty = false; // tracks if the searchbar was empty before the new entry 
 const SearchBar = ({
   manageQuery,
   onSelect,
-  onChange,
   className,
   input_className,
 }: SearchProps) => {
@@ -67,9 +65,6 @@ const SearchBar = ({
   //update url with what's in value
   function updateQueries(newValue: SearchQuery[]) {
     if (typeof manageQuery !== 'undefined' && router.isReady) {
-      if (typeof onChange !== 'undefined') {
-        onChange(newValue);
-      }
       const newQuery = router.query;
       if (newValue.length > 0) {
         newQuery.searchTerms = newValue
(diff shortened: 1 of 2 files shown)


AbhiramTadepalli:
Should the text be "Select a course or professor before searching" instead of "Select an option before searching"

src/pages/index.tsx
```diff
@@ -14,11 +13,8 @@ import searchQueryLabel from '../modules/searchQueryLabel/searchQueryLabel';
  * Returns the home page with Nebula Branding, waved background, and SearchBar Components
  */
 const Home: NextPage = () => {
-  const [errorMessage, setErrorMessage] = useState(false);
   function searchOptionsChange(chosenOptions: SearchQuery[]) {
-    if (chosenOptions.length) {
-      setErrorMessage(false);
-    }
+    // does nothing on change
   }
```
TyHil, line 18: can you remove this function if it does nothing

src/components/common/SearchBar/searchBar.tsx
```diff
@@ -332,15 +334,30 @@ const SearchBar = ({
...
-        Search
-      </Button>
+        <Button
+          variant="contained"
+          disableElevation
+          size="large"
+          className={
+            value.length == 0
+              ? 'shrink-0 normal-case bg-royal hover:bg-royalDark text-cornflower-100'
+              : 'shrink-0 normal-case bg-royal hover:bg-royalDark'
```
TyHil, line 354: can this be:

`'shrink-0 normal-case bg-royal hover:bg-royalDark' + (value.length == 0 ? ' text-cornflower-100' : '')`

so it's more obvious what's changing

src/components/common/SearchBar/searchBar.tsx
```diff
@@ -192,6 +193,7 @@ const SearchBar = ({
 
   //update parent and queries
   function onSelect_internal(newValue: SearchQuery[]) {
+    setErrorTooltip(!newValue.length); //Check if tooltip needs to be displayed
```
TyHil, line 196: the tool tip never goes away on the homepage, not sure why, seems this should do it.

TyHil (changes requested):
The search bar still lets you search with a blank field on the dashboard, should make it not allow you like on the homepage?

Otherwise, left some comments

TyHil:
> Should the text be "Select a course or professor before searching" instead of "Select an option before searching"?

That sounds good, fine with either

AbhiramTadepalli:
> The search bar still lets you search with a blank field on the dashboard, should make it not allow you like on the homepage?
> 
> Otherwise, left some comments

Fixed this and everything else, can you check?

AbhiramTadepalli:
The search button is a _little_ shorter than the searchbar, is that intentional?

TyHil:
Looks all good to me!

> The search button is a _little_ shorter than the searchbar, is that intentional?

Unintentional, just mui default button height

TyHil (approved):

AbhiramTadepalli:
> Unintentional, just mui default button height

should I change?

PR #234, opened 2024-10-07
172 bug properly load grade/rmp data in the search results table

### LINKED ISSUE #172: [Bug]: Some grade/RMP data never loads in the search results table

All possible grade and RMP data is loaded and displayed without extra user input

### What happened?

Grade and RMP data is sometimes loaded but not displayed. I.e. the skeleton wavy loaded never goes away.

This is because the component displaying this data does not realize the data it is displaying has changed status from `loading` to `done`. The component never realizes because it is technically displaying the same object of data, just some of the data in the object has changed. To fix this, every call to `setGrades` and `setRmp` must duplicate the object it takes in (`const newRmp = { ...rmp };
`) and edit/return that duplicated object (`newRmp`) instead of just editing the same object (`rmp`).

### Code of Conduct

AbhiramTadepalli:
## Overview

Resolves #172 by properly triggering the state change from loading to done

## What Changed

Modified setGrades and setRMP in index.tsx so that a newVal was returned instead of the same object, which triggers a state change.

### DIFF
--- a/src/pages/dashboard/index.tsx
+++ b/src/pages/dashboard/index.tsx
@@ -345,7 +345,8 @@ export const Dashboard: NextPage = () => {
     setChosenSessions((old) => {
       const newVal = func(old);
       if (results.state === 'done') {
-        setGrades((grades) => {
+        setGrades((oldGrades) => {
+          const grades = { ...oldGrades }
           //Relavent keys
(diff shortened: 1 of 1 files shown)


TyHil (approved):
Looks great!

Issue #202, opened 2024-09-24
[Feature Request] Expand card if clicked anywhere on the component


AbhiramTadepalli:
# Overview

Since the arrow is a little less noticeable and a smaller touch target, allow users to click anywhere on the search result row to expand the card

To close the card, clicking anywhere on the header of the card should close it.

PR #237, opened 2024-10-07
Card expands when clicked anywhere along the row

AbhiramTadepalli:
## Overview

Closes https://github.com/UTDNebula/utd-trends/issues/202

## What Changed

Added an onClick for the TableRow that triggers when the row is clicked anywhere
**except** when you click on the compare checkbox.

## Other Notes

There is a deadzone between the checkbox and the name label for the result where clicking does not trigger the row expansion.

### DIFF
--- a/src/components/common/SearchResultsTable/searchResultsTable.tsx
+++ b/src/components/common/SearchResultsTable/searchResultsTable.tsx
@@ -110,7 +110,10 @@ function Row({
 
   return (
     <>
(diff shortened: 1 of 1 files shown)


AbhiramTadepalli:
> There is a deadzone between the checkbox and the name label for the result where clicking does not trigger the row expansion.

@TyHil is this acceptable? Would you do this FR any other way?

TyHil:
Yeah seems like a great solution!

TyHil (approved):

PR #240, opened 2024-10-07
chore: added semester dropbox & dropdown menu

sparklingwter:
completed:
* semester dropbox
* dropdown menu with all the academic year options
* select all
* selecting individual works
* select all is chosen at default

will not be completed in this PR:
* recent option does not work


TyHil:
decided to just remove the grade data chip instead of the whole result when filtering to the point of no GPA for a result

AbhiramTadepalli:
> decided to just remove the grade data chip instead of the whole result when filtering to the point of no GPA for a result

I think the result should be removed

AbhiramTadepalli:
Maaybe for deselect all, I like your idea, but if I select 1 semester, every result w/o gpa should disappear

AbhiramTadepalli:
How does this impact compare page?

Whatever the filter is set to impacts ALL the courses in compare.
- [ ] Need to handle blank data in compare (for GPA)

AbhiramTadepalli (approved):
Still some more to do later in #249

PR #243, opened 2024-10-07
Package updates

TyHil:
In an attempt to resolve vulnerabilities before release, I updated all the npm packages using [npm-check](https://www.npmjs.com/package/npm-check). Everything seems to still be working fine, although just a second check through of all functionalities would be good.

This should clear up some of the dependabot alerts once merged in from here: https://github.com/UTDNebula/utd-trends/security/dependabot. If any aren't fixed, I'll go through them and make a quick PR for each vulnerability.


AbhiramTadepalli (approved):
alr man wtvr u say

PR #245, opened 2024-10-08
on a new search changes focus to the first tab (index 0) to show new information

### LINKED ISSUE #210: [Feature Request] When the user adds to compare, switch to the compare tab
# Overview

~Why: So the user can see the new information loaded on a search. Also the compare tab does not have new information during a new search. _This is for a "second search", so after I search for cs 2337, if I search for John Cole, the RHS carousel should default to showing the Professor Overview tab._~

To advertise Compare, and make it easier to use, switch tab to compare if a user adds a course to compare

AbhiramTadepalli:
resolves #210


TyHil (approved):
yay!

PR #248, opened 2024-10-08
Add analytics

TyHil:
Added google analytics based off of this: https://medium.com/readytowork-org/google-analytics-in-next-js-a26cc2b28db5

Can't guarantee it works as it takes some time to show up but we'll see


AbhiramTadepalli (commented):
Datalayer works, so I'm thinking this is working, but need to verify

AbhiramTadepalli (approved):
Go ahead with it

PR #254, opened 2024-10-08
Rmp fix

TyHil:
Noticed the way this runs, when it checks if the returned prof match the given name, it check that the whole "first name" is in the returned name. So if a name is `first middle last`, it would look for `first middle` in the rmp first, and return no data found.


TyHil:
Yes! Ethan found this one: Mehra Borazjany

AbhiramTadepalli (approved):
gotcha 👍

PR #255, opened 2024-10-08
Use @next/third-parties

TyHil:
Change the way we setup Google Analytics per @RubenOlano 's advice to use the Next.js provided way.


AbhiramTadepalli:
How does this change google analytics

TyHil:
I shouldn't, this is just Next.js's suggested way to use google analytics

AbhiramTadepalli (approved):
👍

PR #257, opened 2024-10-09
Pull all colors from tailwind

TyHil:
## Overview

Brand colors were being defined in multiple places which makes any updates difficult.

## What Changed

All the colors are now pulled from the Tailwind config file.


AbhiramTadepalli (approved):
Helps a lot

PR #258, opened 2024-10-14
Remove Vercel Analytics

TyHil:
Not positive we want to do this but since we now have Google Analytics, we can remove Vercel analytics


AbhiramTadepalli:
I want to keep both for a bit and compare which one is better

TyHil:
Just lost Vercel analytics for the next 7 days since its the free tier

AbhiramTadepalli:
Google analytics does pretty much everything except speed insights
Which we don't have to monitor imo it's pretty high

TyHil:
Speed insights is a separate thing in Vercel. This PR keeps speed insights while removing analytics

AbhiramTadepalli (approved):
Bye bye

PR #261, opened 2024-10-15
Change dashboard page name

TyHil:
Wanted to change the dashboard page name so the below metric in Google Analytics would have some meaning. I named it "Dashboard - UTD Trends" but "Results - UTD Trends" might be more appropriate.

image


AbhiramTadepalli:
@TyHil I think "Results - UTD Trends" is the best

TyHil:
@AbhiramTadepalli change made

AbhiramTadepalli (approved):
Great!

Issue #199, opened 2024-09-23
[Feature Request] Resizable Dashboard page


AbhiramTadepalli:
# Overview

Allow users to resize the ratio of left-hand-size to right-hand-size
This should be done by dragging from the border/middle of the LHS and RHS

# Proposal
- [x] Draggable LHS and RHS
- [x] Automatic resizing/reformatting of the components inside LHS/RHS
- [x] Add limits to how small each side can be
- [x] Some visual indicator (probably on hover, the cursor should change) that
- [x] Default to a 50-50 split
- [x] Test for different PC screen sizes

## Supplementary Information
Good luck.

AbhiramTadepalli:
I recommend 2-3 people work on this because of how many tasks there are

PR #264, opened 2024-10-16
#199 resizable dashboard

HershLalwani:
## Overview

Completes #199 

A way for the user to resize the RHS and LHS to improve readability of the compare tab and search list.

## What Changed

Changes in dashboard/index.tsx swapped grid for react-resizable-panels to display the RHS & LHS. There are functions to show the correct UI when accessing from desktop or mobile.

## Other Notes

Several Libraries were attempted, but react-resizable-panels worked best for our needs.


TyHil:
@AbhiramTadepalli double check LHS getting too small and mobile issues
maybe also snapping back to 50/50?

TyHil:
Mobile seems to work on firefox for me but not chrome, may just need to go back to the old Grid based on a useMediaQuery

TyHil:
@AbhiramTadepalli this should be ready now except I can't seem to get the sticky RHS working when merging in develop

TyHil:
@AbhiramTadepalli Ok figured it out. Apparently overflow hidden on any parents screws with position sticky. All ready now.

AbhiramTadepalli:
Okay so every time you add to compare it makes the rhs's minimum width wider. imo minwidth should stay the same because we have horizontal scroll in compare

TyHil:
Idk why setting min width fixes this but it does

AbhiramTadepalli:
ah sorry. Bug:

1) Make LHS bigger, 
2) open a card, 
3) make LHS smaller, 
4) the LHS does not get smaller, RHS just moves on top of it. Expected behavior is LHS resizes (including the contents of the card)

TyHil:
> @TyHil one last question The LHS locks at a min-width. In split-screen (when the browser window is half the size) the RHS keeps going until the edge of the name. Is that the intended behavior?
> 
> In full-screen, the RHS just stops at the LHS's min-width

i dont't quite get this, can you screenshot?



> ah sorry. Bug:
> 
>     1. Make LHS bigger,
> 
>     2. open a card,
> 
>     3. make LHS smaller,
> 
>     4. the LHS does not get smaller, RHS just moves on top of it. Expected behavior is LHS resizes (including the contents of the card)

im getting this behavior like 1 in 10 times, but if you re expand the LHS and re collapse it behaves as intended. is that acceptable?

AbhiramTadepalli:
> > @TyHil one last question The LHS locks at a min-width. In split-screen (when the browser window is half the size) the RHS keeps going until the edge of the name. Is that the intended behavior?
> > In full-screen, the RHS just stops at the LHS's min-width
> 
> i dont't quite get this, can you screenshot?

image
It covers the bubbles, does not make the row skinnier

AbhiramTadepalli:
> im getting this behavior like 1 in 10 times, but if you re expand the LHS and re collapse it behaves as intended. is that acceptable?

Do you know why it's inconsistent? Or the cause?

TyHil:
I have no idea

AbhiramTadepalli:
Since it mostly works, I'll merge it in
- [ ] [Bug]: Resizable dashboard does not shrink an expanded card on the LHS, sometimes

AbhiramTadepalli (approved):
Great work @HershLalwani

Issue #270, opened 2024-10-22
[Bug]: Auto update new courses in autocomplete graph


jpahm:
### What did you expect to happen?

The ability to search for new courses from the current (24f) semester, i.e. ECS 2390.

### What happened?

The course cannot be searched and isn't provided via autocomplete despite the course data being present in the API.

### Version

Beta 0.1.0

### What browsers are you seeing the problem on?

Firefox

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jpahm:
Note that a good long-term solution for this sort of issue will probably involve setting up a system to pull the latest DAG data from the API's `autocomplete` endpoint and generating the associated autocomplete graphs.

TyHil:
Ah I didn't know the autocomplete endpoint was functional, I'll work on that

TyHil:
On hold until the API endpoint is deployed

AbhiramTadepalli:
@TyHil when someone searches for ECS 2390, should we display ECS 3390 distribution data?

TyHil:
Idk, might be a little misleading to display 3390 data for 2390 since there have probably been some class changes. Hopefully people would know to look up 3390, but the fact they have to do that would indicate it is different from 2390.

TyHil:
This issue is blocked by [nebula-api/#225](https://github.com/UTDNebula/nebula-api/issues/225)

> @TyHil when someone searches for ECS 2390, should we display ECS 3390 distribution data? and @jpahm when we call for ECS2390 grade distribution data, does API provide 3390 data from past semesters (also is there 2390 grade data yet)?

This is blocked by [nebula-api/#223](https://github.com/UTDNebula/nebula-api/issues/223)

AbhiramTadepalli:
> Idk, might be a little misleading to display 3390 data for 2390 since there have probably been some class changes. Hopefully people would know to look up 3390, but the fact they have to do that would indicate it is different from 2390.

When https://github.com/UTDNebula/nebula-api/issues/223 is completed, it will in fact group ECS 2390 and ECS 3390.

Now, _we_ could detect the change maybe and let users know the course name has changed? That requires that the course API also groups by `internal_course_number`

TyHil:
If API is doing the grouping, seems reasonable to just do that then

mikehquan19:
[...]

PR #271, opened 2024-10-22
Use new autocomplete data from API

TyHil:
Temp fix for #270 , rebuilt the autocomplete graph and combos table from new data Josh sent over


AbhiramTadepalli (approved):
ty

PR #277, opened 2024-10-25
Did the wrong name for the dashboard

TyHil:
In #261 changed the dashboard name to "Results - Nebula Labs" instead of "Results - UTD Trends".
Oops


AbhiramTadepalli (approved):
Ah I should've caught that.

PR #291, opened 2024-10-30
Set cursor pointer

TyHil:
> On hovering on the LHS cards, the cursor behavior should be a pointer. This is a css one-liner (cursor: pointer) and is very intuitive.
- @Segneur

### DIFF
--- a/src/components/common/SearchResultsTable/searchResultsTable.tsx
+++ b/src/components/common/SearchResultsTable/searchResultsTable.tsx
@@ -134,18 +134,15 @@ function Row({
             </IconButton>
           </Tooltip>
         </TableCell>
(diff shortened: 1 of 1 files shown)


AbhiramTadepalli:
@TyHil hover over the name makes it a text cursor

TyHil:
Yeah I thought it made more sense if you wanted to copy the prof name, actually let me make is so that clicking on the next doesn't open the row, so you can double click to copy without opening and closing the row

TyHil:
@AbhiramTadepalli check it out now

AbhiramTadepalli (approved):
Mostly good. #296 should fix it so that everything but the actual width of the text is expandable

PR #292, opened 2024-10-30
288 bug handle null rmp data

### LINKED ISSUE #288: [Bug]: LHS & RHS handle null RMP data

If there is no RMP data, it should be indicated properly (https://trends.utdnebula.com/dashboard?searchTerms=Yu+Chung+Ng)

### What happened?

I get 0 as the rating (when it should be null) and -1% for would take again.
image

Same applies for the rating component, hovers on LHS, and RHS compare
image

### Version

trends.utdnebula.com

_No response_

### Relevant log output

_No response_

### Code of Conduct

George-Zhang5246:
## Overview

Resolves #288 

## What Changed

ratemyprofessorScraper now gives an error when the rmp data contains 0 ratings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# UTD-Trends
+# UTD Trends
 
-_UTD-Trends a data visualization tool built to help students view historical course and section data using the Nebula API_
+_Easy access to all the information you need to plan your schedule_
 
 [![Commitizen friendly](https://img.shields.io/badge/commitizen-friendly-brightgreen.svg)](http://commitizen.github.io/cz-cli/)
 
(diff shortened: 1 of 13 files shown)


AbhiramTadepalli:
@TyHil in relation to our other convo about no grade data, should we show some message that this prof has no RMP data?

TyHil:
Hmm yeah I suppose it should maybe just say reviews: 0 and have the link to the RMP page

George-Zhang5246:
I guess I can do that, though it would need modifications to all of the components as opposed to just the RMP scraper.

AbhiramTadepalli (approved):
Great job @George-Zhang5246!

TyHil (approved):

PR #296, opened 2024-10-31
Fix loading jumping

TyHil:
## Overview

The UI jumps around a bit as it transitions through the 2 loading states to the final done state.

## What Changed

This fixes that by making the name take up all remaining width in the table so the placement of the last 2 columns is not dependent on the longest professor name.

### DIFF
--- a/src/components/common/SearchResultsTable/searchResultsTable.tsx
+++ b/src/components/common/SearchResultsTable/searchResultsTable.tsx
@@ -58,18 +58,21 @@ function LoadingRow() {
       <TableCell>
         <Checkbox disabled />
       </TableCell>
-      <TableCell component="th" scope="row">
-        <Typography className="w-[20ch] leading-tight text-lg text-gray-600 dark:text-gray-200">
+      <TableCell component="th" scope="row" className="w-full">
+        <Typography className="w-full leading-tight text-lg">
           <Skeleton />
         </Typography>
       </TableCell>
-      <TableCell align="right">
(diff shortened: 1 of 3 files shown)


AbhiramTadepalli:
@TyHil commit the fixes to #291

TyHil:
Give it a check now

src/components/common/SearchResultsTable/searchResultsTable.tsx
```diff
@@ -58,18 +58,21 @@ function LoadingRow() {
...
       </TableCell>
-      <TableCell align="right">
-        <Skeleton variant="rounded" className="rounded-full px-5 py-2 ml-auto">
-          <Typography className="text-base">4.00</Typography>
+      <TableCell align="center">
+        <Skeleton
+          variant="rounded"
+          className="rounded-full px-5 py-2 w-16 block mx-auto"
+        >
+          <Typography className="text-base w-6">A+</Typography>
```
AbhiramTadepalli, line 71: @TyHil should this be w-16? idk if it functionally matters but j wondering
Otherwise all good

AbhiramTadepalli (changes requested):

src/components/common/SearchResultsTable/searchResultsTable.tsx
TyHil: The text needed to be w-6 to make the whole thing w-16 with padding. Was kind of a weird way to do it tho, just pushed a better way.

AbhiramTadepalli (approved):
¡magnífico!

PR #307, opened 2024-11-06
Separate names in autocomplete graph

### LINKED ISSUE #260: [Bug]: Professors with middle names can be hard to find in the search bar

Typing Mehra Borazjany into Trends should pull up the correct professor but as soon as you type the B, the option goes away.

### What happened?

Mehra Borazjany's full name is Mehra Nouroz Borazjany and is added to the autocomplete graph from the below data:
```
{
  "first_name": "Mehra Nouroz",
  "last_name": "Borazjany"
}
```

So the graph builds like so:
```
Start --> Mehra Nouroz ↘
      ↘-----------------> Borazjany --> data results
```

When it should build like so:
```
Start --> Mehra -----------------↓
      ↘         ↘---> Nouroz ↘  ↓
        ↘---------------------> Borazjany --> data results
```

This allows typing Mehra B and bypassing Nouroz to get to the data results so you can type the name wither with or without the middle name.

This should be possible by modfiying the `addProfessor` function in `generateAutocompleteGraph.ts` to split the first name by spaces and process each name.

### Version

Beta 0.1.0

_No response_

### Relevant log output

_No response_

### Code of Conduct

TyHil:
## Overview

Resolves #260 

Professors with 3+ names are hard to search for when starting with first name

## What Changed

Instead of treating multiple first names as a single node in the autocomplete graph, this spilts each name onto its own node that can be bypassed to get to the last name and thus professor data.


AbhiramTadepalli (approved):
Works!

PR #325, opened 2024-11-20
RHS compare indicators and resizing when no semesters selected

### LINKED ISSUE #304: [Bug]: RHS Compare Skeleton sizing
https://github.com/UTDNebula/utd-trends/commit/af2cd287f9d4f0a5ce79c88695b694988c350248 - when you unselect "select all" the gpa and bar graphs rightly disappear (CS 1200 Gordon Arnold & CS 1200 Wei-Pang Chin in compare), but the components move a lot (I'd like it to be stationary).

Just modify the skeleton

egsch:
## Overview
Resolves #304 

## What Changed
Re-adds check for -1 GPA (for some reason it was showing -1 as the GPA again on some professors). Adds empty space with same dimensions in the compare table when there is no GPA is -1, which minimizes jumping.

## Other Notes
The bar graph still moves around a bit because the labels disappear and the graph height expands. If we wanted we could keep the labels visible to avoid this, but it seems like that could be confusing and it would get rid of the empty state text.

### DIFF
--- a/src/components/common/SingleProfInfo/singleProfInfo.tsx
+++ b/src/components/common/SingleProfInfo/singleProfInfo.tsx
@@ -116,11 +116,20 @@ function SingleProfInfo({ rmp }: Props) {
   return (
     <Grid container spacing={2} className="p-4">
       <Grid size={6}>
-        <p className="text-xl font-bold">{rmp.data.avgRating}</p>
+        <p className="text-xl font-bold">
+          {typeof rmp.data.avgRating !== undefined && rmp.data.avgRating > 0
+            ? rmp.data.avgRating
+            : 'N/A'}
+        </p>
         <p>Professor rating</p>
       </Grid>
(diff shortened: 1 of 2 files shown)


AbhiramTadepalli (approved):
Great solution!

AbhiramTadepalli:
@egsch @TyHil slightly unrelated but what do you think about the commit I made? For example, Alexander Cornelius does not have a WouldTakeAgain metric, so I think we should just show "N/A". There is already a check for if numRatings is 0.

TyHil:
Oh yeah seems resonable

egsch:
> @egsch @TyHil slightly unrelated but what do you think about the commit I made? For example, Alexander Cornelius does not have a WouldTakeAgain metric, so I think we should just show "N/A". There is already a check for if numRatings is 0.

That looks good!

PR #327, opened 2024-11-20
Fix for UTD Grades Ad covering searchbar on android firefox when keyboard is open

TyHil:


TyHil:
virtual keyboard behavior seems to be pretty different across browsers

PR #329, opened 2024-11-22
Remove styleguide

TyHil:
## Overview

I don't think anyone knows how to use the `react-styleguidist` stuff. I was added in a couple commits in 2022 and hasn't been touched since then. The main benefit I see of it is it allows README documents to render components they are documenting which would be nice.

## What Changed

Removed the `styleguide` and `styleguidistDocs` folders, the `styleguide.config.js` file, and a bunch of packages in `package.json`. I'm not sure how many of the packages being removed were needed for `react-styleguidist` and how many were just unused.


TyHil:
@NishilJ if you have any luck getting `react-styleguidist` to work we can discard this PR.

AbhiramTadepalli:
@NishilJ do you think we need StyleGuides or should we remove em?

AbhiramTadepalli:
@NishilJ we're keeping em right?

NishilJ:
I think it is valuable for documentation, if you think so, my pr would replace this one.

PR #332, opened 2024-11-24
Fix load on enter

AbhiramTadepalli:
## Overview

After https://github.com/UTDNebula/utd-trends/pull/298/files, since searching is initiated quicker, user feels some lag while waiting for search results

## What Changed
Set the loading state quicker, and add a loading icon in the search button

### DIFF
--- a/src/components/common/TableSortLabel/tableSortLabel.tsx
+++ b/src/components/common/TableSortLabel/tableSortLabel.tsx
@@ -1,28 +1,13 @@
-import {
-  styled,
-  TableSortLabel as _TableSortLabel,
-  type TableSortLabelProps,
-  useMediaQuery,
-} from '@mui/material';
-import React from 'react';
+import { styled, TableSortLabel } from '@mui/material';
 
-function TableSortLabel(props: TableSortLabelProps) {
-  //Selected arrow color
-  const sortArrowColor = useMediaQuery('(prefers-color-scheme: dark)')
-    ? 'white'
-    : 'black';
+//Change arrow color for TableSortLabel
+const StyledTableSortLabel = styled(TableSortLabel)({
+  '& .MuiTableSortLabel-icon': {
+    opacity: 0.5, // Ensure the arrow is always visible
+  },
+  '&.Mui-active .MuiTableSortLabel-icon': {
+    color: 'var(--mui-palette-text-primary)', // Brighten the arrow
+  },
+});
 
-  //Change arrow color for TableSortLabel
-  const StyledTableSortLabel = styled(_TableSortLabel)({
-    '& .MuiTableSortLabel-icon': {
-      opacity: 0.5, // Ensure the arrow is always visible
-    },
-    '&.Mui-active .MuiTableSortLabel-icon': {
-      color: sortArrowColor, // Brighten the arrow
-    },
-  });
-
-  return <StyledTableSortLabel {...props} />;
-}
-
-export default TableSortLabel;
+export default StyledTableSortLabel;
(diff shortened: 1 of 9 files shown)


AbhiramTadepalli:
I am calling setSearchResultsLoading() inside the useEffect that has searchResultsLoading as a dependency. Is this bad practice?

AbhiramTadepalli:
I don't really like the solution. It works but it introduces an unnecessary state I feel like combining it with results.state is probable better but idk how to fully.

AbhiramTadepalli:
Another problem:

On search from landing page to dashboard, it calls DashboardEmpty and then DashboardError before the searchResultsTable Skeletons

If I modify: 
```if (results.state === 'done' && courses.length === 0 && professors.length === 0) {
    contentComponent = <DashboardEmpty />;```
and Remove the DashboardError elseif, then for a second, it shows the "No Results" state from searchResultsTable.tsx

Ideally:
Initiate search from landing -> directly shows srt's loading state, nothing in between

TyHil:
Can revert but lmk what you think of these changes to use `results.state` and remove the flashing `DashboardError` and `DashboardEmpty`.

Opinion: I don't feel the spinner in the search bar is all that necessary when the whole page changes to a loading state.

Current issue (possibly unsolvable): There's also an issue when reloading/opening the dashboard (but _not_ when navigating to it from the homepage) where the `searchResultsTable` and RHS cards flash white. I seems this is just in the split second before CSS is loaded but it is odd

TyHil:
> the topMenu does still jump around because SSR seems to be rendering it as mobile first because of the media query it uses. there might be a css solution

solved! turns out using an actual media query instead of useMediaQuery fixes this. Probably because SSR can guess what the size should be

AbhiramTadepalli:
- [ ] Do not search when user clicks on an autocomplete suggestion
- [ ] If you type out cs 1200, space, cs2337, space, clicking enter does not search

AbhiramTadepalli:
Everything else works perfectly. Searches appear faster because there is less lag in between and code-wise it's very simple

TyHil:
> Do not search when user clicks on an autocomplete suggestion
> If you type out cs 1200, space, cs2337, space, clicking enter does not search

I think #298 caused both of these issues as they are present on that PR, dev, and prod.

AbhiramTadepalli (commented):
nice! So are all the breakpoints local to each component now?

AbhiramTadepalli:
> > Do not search when user clicks on an autocomplete suggestion
On the landing page this works as intended. The problem is only with the searchbar on dashboard

AbhiramTadepalli:
searchbar.tsx is very convoluted

AbhiramTadepalli:
> > If you type out cs 1200, space, cs2337, space, clicking enter does not search
> 
> I think #298 caused both of these issues as they are present on that PR, dev, and prod.

Fixed this, you were right @TyHil

AbhiramTadepalli:
> [ ] Do not search when user clicks on an autocomplete suggestion

For this, I think we should leave the current behavior as-is.

AbhiramTadepalli:
There is some issue when I click enter twice for the same search terms without change, it just perpetually loads

AbhiramTadepalli:
My fix prevents the search when the terms haven't changed. No need for another batch of API calls imo (and if they really want, they can refresh).

Also I'd have to change a few things so that the useEffect triggers properly and all that. Don't think it is needed.

AbhiramTadepalli:
Searchbar action flows:
// updateValue -> onSelect_internal -> updateQueries - clicking enter on an autocomplete suggestion in topMenu Searchbar
  // updateValue -> onSelect_internal -> onSelect (custom function) - clicking enter on an autocomplete suggestion in home page SearchBar
  // params.inputProps.onKeyDown -> handleKeyDown -> onSelect_internal -> updateQueries/onSelect - clicking enter in the SearchBar
  // Button onClick -> onSelect_internal -> updateQueries/onSelect - Pressing the "Search" Button
  
  @TyHil if you want me to put the comments somewhere else instead lmk.

AbhiramTadepalli:
Can you check if the spinner is colored correctly in dark mode? It was before idr which commit changed it

PR #334, opened 2024-11-25
Collapsible Mobile RHS

### LINKED ISSUE #274: [Feature Request] Mobile UI
Search results need to be more easily accessible. Perhaps fit the RHS into a collapsing/expandable element at the top?

TyHil:
Resolves #274


AbhiramTadepalli (approved):
So `open` is always true when the screen is wide?

Do you have to change useMediaQuery?

TyHil:
> So `open` is always true when the screen is wide?

Yes

> Do you have to change useMediaQuery?

Yeah I should do that

TyHil:
I think I'll need #332 merged in to see how these work together before I can get rid of the second useMediaQuery

AbhiramTadepalli:
> I think I'll need #332 merged in to see how these work together before I can get rid of the second useMediaQuery

did you do this?

TyHil:
yeah, as best i could. can't do it all with css so i did still need the useMediaQuery. i just had it start out closed so it looks fine on mobile and expands on desktop into empty space which is fine.

PR #336, opened 2024-12-05
Compare Art

TyHil:
I thought this looked cute. Check out the compare tab when no results are added to it


AbhiramTadepalli:
Maybe different art?

TyHil:
Fair. I just thought the compare tab looked weird so short. Not really sure on what different art would be

TyHil:
:(

PR #339, opened 2025-01-06
Fix loading on homepage & loading spinner color visibility

AbhiramTadepalli:
Made the loading spinner color the same across light and dark mode since the button stays the same color too.

---

Added loading on the homepage with a dummy state that starts at 'done' and is passed into SearchBar (which then sets state to 'loading')

DLC for #332


AbhiramTadepalli:
Is the dummy state in pages/index.tsx good practice?

TyHil (approved):

PR #342, opened 2025-01-14
Swap Mean GPA for Median GPA

AbhiramTadepalli:
## Overview
Median might be more representative of the distribution than mean, since median is resistant to outliers.

## Other Notes

**However, there is still a lot of discussion to be had because this will change how we present & how users perceive courses.**


AbhiramTadepalli:
@TyHil what do you think about using median? For (ie) CS 2337, the median reported seems to represent the distribution a _lot_ better imo.

---
##Some  considerations
- Median GPAs are boring (All medians of a B will be 3.33 always)
- What should be shown on hover?
- Inside the card, can we show mean and median both? Or just mean? Or do away with mean completely?
- What reasons do we have to _not_ switch to median as the average statistic? I think across the board it has been boosting the average compared to mean...

TyHil:
Tbh my statistics knowledge is not up to par enough to know which to use. From basic googling it seems median is for skewed data which ours is likely to be so that seems like a reasonable switch.

Maybe keep average on hover ig.
We can add the median on the expanded card as a bar graph annotation instead of more text below the chart.

PR #343, opened 2025-01-14
Disable Prettier new line changes

TyHil:
Stops prettier from changing all line endings to `\n` which looks like a lot of changes, only for git to remove the changes when you do `git add .`


AbhiramTadepalli (approved):
Simple enough 🙂

PR #358, opened 2025-02-24
263 use grade highlight colors

### LINKED ISSUE #263: [Feature Request] Use grade highlight colors on compare
# Overview

We have 2 arrays of colors we use on the LHS, depending on whether the user is on dark or light mode. However, the compare component uses a dynamic range of colors from #00FF00 to #FF0000.

# Proposal

Update the compare component to use the dark/light arrays of colors. Either just do a gradient between the first and last color or somehow use all the colors in the array.

rchowdam1:
Resolves #263 

Use the functionality that searchResultsTable.tsx (gpaToColor()) uses to pick colors for the class grade into compareTable. If the row in compare table is the GPA, then use gpaToColor(). All other rows are the same as before.

### DIFF
--- a/src/components/compare/CompareTable/compareTable.tsx
+++ b/src/components/compare/CompareTable/compareTable.tsx
@@ -190,9 +190,7 @@ function GradeOrRmpRow<T>({
                           : colorMidpoint(
                               goodValue,
                               badValue,
-                              name === 'GPA'
-                                ? getValue(value.data)
-                                : getValue(value.data),
+                              getValue(value.data),
                               rainbowColors[12],
                               rainbowColors[0],
                             ),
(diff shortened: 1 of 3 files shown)


src/components/compare/CompareTable/compareTable.tsx
```diff
@@ -152,11 +184,18 @@ function GradeOrRmpRow<T>({
...
+                          : colorMidpoint(
+                              goodValue,
+                              badValue,
+                              name === 'GPA'
+                                ? getValue(value.data)
+                                : getValue(value.data),
+                              rainbowColors[12],
+                              rainbowColors[0],
+                            ),
                     }}
```
AbhiramTadepalli, line 199: Can you explain what this does?

rchowdam1: The first check is to call gpaToColor() if the row corresponds to GPA
the second check doesn't actually do anything, forgot to remove it

TyHil:
just moved that function to a module, all looks good to me @AbhiramTadepalli if you'll give a quick check and merge

AbhiramTadepalli (approved):
Great work!

PR #364, opened 2025-02-27
MyPlanner Template

AbhiramTadepalli:
…r page

## Overview

List the GitHub issues containing the issues relevant to this pull request.

[For example, "Resolves #1, closes #2, fixes #3", etc.]

Describe your changes here at a high level, describing how this PR fits into the
rest of the project.

## What Changed

Go more into detail about key files that were modified and why they were
updated. Do not list every file that was modified; only note the ones most
relevant to this feature or bug fix.

## Other Notes

If were roadblocks encountered during development that remain unresolved or any
future additions or changes to make, note them here. Otherwise. Feel free to
delete this section if it isn't needed.

PR #373, opened 2025-02-28
Store and fetch in _app

TyHil:
This is the same code from #364 just without the planner additions. It's just a lot of changes that I'd like to get into develop so there are fewer conflicts when we eventually merge in `my-planner-development`.


AbhiramTadepalli (approved):
Yes!!!

PR #375, opened 2025-03-03
Prevent Users from opening course cards with empty search results

Evelas78:
Used skedge's code to implement canOpen checks if search results didnt exist, disabling the arrow icon if it isnt openable.

### DIFF
--- a/src/components/search/SearchResultsTable/searchResultsTable.tsx
+++ b/src/components/search/SearchResultsTable/searchResultsTable.tsx
@@ -83,13 +83,15 @@ function Row({
   color,
 }: RowProps) {
   const [open, setOpen] = useState(false);
-
+  const canOpen = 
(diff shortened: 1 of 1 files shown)


Evelas78:
> If you want to do this on a branch instead of a fork it might be a bit easier but other than that this is looking good

Ah gotchu. Honestly, this is my first pull request like ever so I'm not really sure how to do that, so for future reference, I'd love a quick explanation how to if possible!

TyHil:
Yeah for sure, will you be at the Trends meeting today? We can work through it there

Evelas78:
> Yeah for sure, will you be at the Trends meeting today? We can work through it there

Yea I will be, sounds good

AbhiramTadepalli (approved):
Looks good to me!

PR #340, opened 2025-01-13
280 feature request indicate most recent semester taught

AbhiramTadepalli:
## Overview

When scanning search results, some professors may have not taught in a while (making them less viable options)

Implements #280 

## What Changed

Added a `most_recent_semester` prop to the `GradesType` type, which is filled in `calculateGrades` (_dashboard/index.tsx_)
Then, call that prop to fill the badge in _searchResultsTable.tsx_'s row

## Other Notes

Issue #280, opened 2024-10-26
[Feature Request] Indicate Most Recent Semester Taught


AbhiramTadepalli:
# Overview

Trends does not indicate the most recent semester a professor has taught. The only way to find this information is by trying out the semester filters. This FR will add text that displays the most recent semester the professor has taught.

AbhiramTadepalli:
> I think we should solve this issue on another branch. This issue doesn't implement the UI we talked about and we're going to process the data like in #365. So this will be good for reference but I'm pretty sure the issue needs to be started from scratch.

_Originally posted by @TyHil in https://github.com/UTDNebula/utd-trends/issues/340#issuecomment-2704726829_

AbhiramTadepalli:
Oh hold until tyler merges the #365 changes into develop

TyHil:
> Oh hold until tyler merges the [#365](https://github.com/UTDNebula/utd-trends/issues/365) changes into develop

just finished that!

AbhiramTadepalli:
Right now I thikn with a mix of recent sem filter, line graph, and "profs teaching next sem", this is not absolutely necessary. We never formally decided, so up to you but this is what i think

PR #376, opened 2025-03-03
365 feature request pass section data for selected results into my planner

### LINKED ISSUE #365: [Feature Request] Pass Section data for selected results into My Planner
# Overview
Once a user clicks to add a search result to compare, pass that data in.

TyHil:
## Overview

Resolves #365 

Adds fetching logic to get sections being taught for course professor combos

## What Changed

Adds `src/modules/useSectionsStore/useSectionsStore.ts` similar to the grades and rmp ones. Differences are this one fetches a common class on initialization (GOVT 2306) to find the newest semester in the database. And for a combo this must do 2 fetches (which is really even more due to pagination) and then take the intersection of them.

## Other Notes

Stole some code in `src/modules/SearchQuery/SearchQuery.tsx` and `src\modules\semesters\semesters.tsx` from #340 which could probably also use this code on the dashboard.

To test, go to the dashboard and add some combos to planner (maybe make sure that prof is actually teaching this semester). Then go to the planner page and look in the console where the fetched grade, rmp, and section data is.

### DIFF
--- a/src/modules/useSectionsStore/useSectionsStore.ts
+++ b/src/modules/useSectionsStore/useSectionsStore.ts
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 
 import fetchWithCache, {
   cacheIndexNebula,
@@ -11,6 +11,7 @@ import {
   type SearchQuery,
   searchQueryLabel,
 } from '@/modules/SearchQuery/SearchQuery';
+import { compareSemesters } from '@/modules/semesters/semesters';
 import type { SectionData } from '@/pages/api/sections';
 
 //Fetch section data from nebula api
@@ -45,6 +46,18 @@ function fetchSectionsData(
   });
 }
 
+// Finding the most recent semester in the database by the newest GOVT 2306 semester (don't judge me)
+function checkLatestSemester() {
+  return fetchSectionsData(
+    { prefix: 'GOVT', number: '2306' },
+    new AbortController(),
+  ).then((res: SectionData) => {
+    return res.sort((a, b) =>
+      compareSemesters(b.academic_session.name, a.academic_session.name),
+    )[0].academic_session.name;
+  });
+}
+
 //Limit cached number of data entries
 const MAX_ENTRIES = 1000;
 
@@ -58,6 +71,12 @@ export default function useSectionsStore(): [
   (combo: SearchQuery, controller: AbortController) => void,
 ] {
   const [sections, setSections] = useState<Sections>({});
+  const latestSemester = useRef('');
+  useEffect(() => {
+    checkLatestSemester().then((result: string) => {
+      latestSemester.current = result;
+    });
+  }, []);
 
   function addToSections(key: string, value: GenericFetchedData<SectionData>) {
     setSections((old) => {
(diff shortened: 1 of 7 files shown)


TyHil:
Just made a change that when this gets sections instead of keeping just the ones for the latest semester (current or upcoming), it stores them like so:

```
export type SectionsType = {
  all: SectionsData;
  latest: SectionsData;
};
```

This should allow #340 for #280 to run `fetchAndStoreSectionsData` for each combo and tell:

- new prof: latest.length == all.length
- not teaching: !latest.length

Or probably better to call `fetchAndStoreSectionsData` for each course query and filter by professor IDs to reduce calls.

src/modules/useSectionsStore/useSectionsStore.ts
```diff
@@ -0,0 +1,152 @@
...
+  ).then((response) => {
+    if (response.message !== 'success') {
+      throw new Error(response.message);
+    }
+    return response.data;
+  });
+}
+
+// Finding the most recent semester in the database by the newest GOVT 2306 semester (don't judge me)
+function checkLatestSemester() {
```
AbhiramTadepalli, line 51: Do you need to do this?
iirc I basically j sorted this and found the latest semester, compared it to currentSem (today)

Just so we don't have to rely on a course?

src/modules/useSectionsStore/useSectionsStore.ts
```diff
@@ -0,0 +1,152 @@
...
+      convertToProfOnly(combo),
+    ]
+      //Remove empty objects (for non combos)
+      .filter((obj) => Object.keys(obj).length !== 0);
+    //Call each
+    Promise.all(
+      seperatedCombo.map((query) => fetchSectionsData(query, controller)),
+    )
+      .then((res: SectionsData[]) => {
+        //Find intersection of all course sections and all professor sections
```
AbhiramTadepalli, line 122: so you get all the sections of the profs, all the sections of the courses, intersect them -> 1 array with all the sections that match the SearchQueries which is then filtered to the latest semester?

src/pages/api/sections.ts
```diff
@@ -0,0 +1,115 @@
...
+  if (typeof url === 'undefined') {
+    res.status(400).json({ message: 'Incorrect query present' });
+    return;
+  }
+  const headers = {
+    'x-api-key': API_KEY,
+    Accept: 'application/json',
+  };
+
+  function recursiveFetch(url: URL, offset: number): Promise<SectionsData> {
```
AbhiramTadepalli, line 79: what does recursiveFetch do?

AbhiramTadepalli (approved):

src/modules/useSectionsStore/useSectionsStore.ts
TyHil: I don't think currentSemester works because this needs to update as soon as new sections are in the API, not based on a month.

src/modules/useSectionsStore/useSectionsStore.ts
TyHil: yes

src/pages/api/sections.ts
TyHil: The API only returns 20 sections at a time so we have to do a few calls to get them all.

src/modules/useSectionsStore/useSectionsStore.ts
AbhiramTadepalli: we know gov is offered ev sem right? wb summer?

src/pages/api/sections.ts
AbhiramTadepalli: Is that a Josh limit or UTD-imposed limit?

src/modules/useSectionsStore/useSectionsStore.ts
TyHil: I wasn't expecting this to work for the summer since I think summer courses are releases at the same time as Spring ones, so we'd need a UI to select between them.

src/pages/api/sections.ts
TyHil: It's a Josh limit

src/modules/useSectionsStore/useSectionsStore.ts
AbhiramTadepalli: summer releases a week before

Actually it's fine I think we shouldn't add a toggle

PR #381, opened 2025-03-04
Change SearchBar search button color

TyHil:
Make SearchBar search button color follow the new main lighter purple color theme


AbhiramTadepalli (approved):
majestic

PR #382, opened 2025-03-04
Open card when clicking on text

TyHil:
I had disabled opening the card when clicking on the text in case you wanted to copy something but this causes too much confusion as the text is the main space to click on the card.

This PR makes clicking on the text open the card.

### DIFF
--- a/src/components/search/SearchResultsTable/SearchResultsTable.tsx
+++ b/src/components/search/SearchResultsTable/SearchResultsTable.tsx
@@ -104,12 +104,7 @@ function Row({
       }
       placement="top"
     >
-      <Typography
-        onClick={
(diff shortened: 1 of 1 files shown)


TyHil:
@Segneur do you think I should change the cursor to `pointer` or leave it as `text`?

Segneur:
> @Segneur do you think I should change the cursor to `pointer` or leave it as `text`?

I think pointer is most intuitive to let people know it can be clicked; any reason you would think to keep it as text?

AbhiramTadepalli (approved):

AbhiramTadepalli:
> > @Segneur do you think I should change the cursor to `pointer` or leave it as `text`?
> 
> I think pointer is most intuitive to let people know it can be clicked; any reason you would think to keep it as text?

Well it is select-able text

PR #387, opened 2025-03-10
Auto-switch to compare on change in compare.length

AbhiramTadepalli:
## Overview

Implements #210 

Idea is so that (a) people find the compare tab (b) if you're adding to compare, you probably want to see it next


AbhiramTadepalli:
ready to review @TyHil

PR #319, opened 2024-11-16
Tutorial

TyHil:
## Overview

Ended up going with a step by step tutorial since I couldn't figure out how to have all the popups on screen at once, and that wouldn't work on mobile anyway.

## What Changed

Added a tutorial component and various tags for elements it highlights sprinkled around the code.

## Other Notes

I have not added a feedback popup. This may require more discussion but making it so easy to submit feedback could open us up to spam. Currently, technical knowledge is required to spam us but this could make it too easy. I could see it working if it went to a different place and had just the question with no star rating. But overall I don't see the point in asking the people who have actually hit the question mark icon what they think about the site when they are the ones who have barely used it.

There is a known issue on mobile where each step in the tutorial rescrolls from the top of the page to the designated element, not sure why rn.

PR #388, opened 2025-03-10
313 add a what's new component

ridham15:
Added a new "Whats New" button in the topmenu that displays the latest release from the UTD-Trends release page. You can click on the message and it will directly take you to the page of that particular release. I also used local storage to mark messages as read/unread based on if the user is seeing them for the first time or not. Let me know if any changes need to be made. 

Issues:
I am not able to fix the card from not moving when you scoll so let me know if you have any ideas on how I can do it.


TyHil (dismissed):
This looks good. Couple suggestions:

- Can we move this to a separate component so it doesn't crowd TopMenu
- Using a MUI Popover element will work better that setting the `left` style as it will update when the window resizes
- Can we style it like the feedback popup or change the feedback popup so they're the same? @AbhiramTadepalli 

There is similar code in the tutorial PR (#319) if that's any help

ridham15:
I made a new folder for the WhatsNew component. Is there a particular place you would like this to be in? Currently its inside the navigation folder.

TyHil:
id say `common` makes more sense

AbhiramTadepalli:
> Can we style it like the feedback popup or change the feedback popup so they're the same? @AbhiramTadepalli

dyk why `dark:bg-haiti` is ```#242643``` in the feedbackpopup but ```#090b2c``` when I use it in whatsnew?

TyHil:
I think the lack of an `elevation={0}` prop on the `SnackbarContent` component is adding a `background-image` overlay on the component that changes the color.

AbhiramTadepalli:
@TyHil @ridham15 should we have the 2 most recent releases instead of 1? That way people can see both like Line Graph as a release and MyPlanner.

Perfectly fine if not, just wanted to ask what y'all think

AbhiramTadepalli (approved):
Great job @ridham15 !

ridham15:
@AbhiramTadepalli We can definetly do that since it won't occupy much space on the page and since line graph is a new feature that we just added it would be better if we can have hint it in multiple places.

TyHil:
@AbhiramTadepalli can you use the elevation prop instead of hard coding the color

AbhiramTadepalli:
> @AbhiramTadepalli can you use the elevation prop instead of hard coding the color

Done, don't understand how it works

PR #389, opened 2025-03-12
Added course Filter

Matrixk1ng:
Added Course filter. Not finished data works just Ui

## Overview

Added A course filter for Trends

[For example, "Resolves #1, closes #2, fixes #3", etc.]

Changed the grades.ts api and added the sectionType api endpoint, modified index.ts to accomodate the data.

## What Changed

## Other Notes

The data works and changes with the filter properly, but the Ui does not change not done

### DIFF
--- a/src/components/search/Filters/Filters.tsx
+++ b/src/components/search/Filters/Filters.tsx
@@ -17,11 +17,15 @@ import gpaToLetterGrade from '@/modules/gpaToLetterGrade/gpaToLetterGrade';
 const minGPAs = ['3.67', '3.33', '3', '2.67', '2.33', '2'];
 const minRatings = ['4.5', '4', '3.5', '3', '2.5', '2', '1.5', '1', '0.5'];
 
+
 interface FiltersProps {
   manageQuery?: boolean;
   academicSessions: string[];
   chosenSessions: string[];
   addChosenSessions: (arg0: (arg0: string[]) => string[]) => void;
+  courseType: string[];
+  addChosenCourseType: (arg0: (arg0: string[]) => string[]) => void;
+  chosenCourseType: string[]
 }
 
 /**
@@ -32,6 +36,9 @@ const Filters = ({
   academicSessions,
   chosenSessions,
   addChosenSessions,
+  courseType,
+  addChosenCourseType,
+  chosenCourseType
 }: FiltersProps) => {
   const [minGPA, setMinGPA] = useState('');
   const [minRating, setMinRating] = useState('');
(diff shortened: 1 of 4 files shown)


src/components/search/Filters/Filters.tsx
```diff
@@ -130,6 +156,7 @@ const Filters = ({
       return 0;
     } else return x;
   }
+  const [trigger, setTrigger] = useState(0);
```
TyHil, line 159: is this trigger necessary?

src/components/search/Filters/Filters.tsx
```diff
@@ -314,6 +342,77 @@ const Filters = ({
           </Select>
         </FormControl>
       </Tooltip>
+      {/* Course Type dropdown */}
+
+      <Tooltip title={'Select Course Type to choose from'} placement="top">
+        <FormControl
+          size="small"
+          className={`w-full ${
+            chosenSessions.length !== academicSessions.length
```
TyHil, line 351: this should be course type

src/pages/dashboard/index.tsx
```diff
@@ -305,9 +314,24 @@ export const Dashboard: NextPage<Props> = (props: Props): React.ReactNode => {
...
+        );
+      }
+      recalcAllCompareGrades(compare, newVal);
+      return newVal;
+    });
+  }
 
   //Add a set of sessions to the compiled list, removing duplicates and keeping a sorted order
-  function addAcademicSessions(sessions: string[]) {
+  function addAcademicSessions(sessions: string[], courseTypes: string[]) {
```
TyHil, line 334: i don't this function should need the course types, or have any effect on them

src/components/search/Filters/Filters.tsx
TyHil: looks like it still uses chosenSessions

TyHil (commented):
any idea what the xUx section type is? shows up on cs 2340

Matrixk1ng:
for xUx, i have no idea i tried looking it up earlier, did not find much

TyHil:
ok, fair enough

TyHil:
figured out they're summer sections!

Matrixk1ng:
k

AbhiramTadepalli:
Bugs I noticed
1. In compare, search john cole, add CS 1200. Unselect all and then select honors. The GPA changes, but the graph is broken. (It works when you manually unselect everything except honors)
2. Select Night Lecture only for JC. Open his overall course -- graph does not populate. When you click on it to close, the graph populates rq before the card closes
3. Line graph is jacked up
4. With a filter on, add a few to the compare graph, then take all but 1 off: 
image
_Didn't know blud could do that lowk impressed_
5. @TyHil did you add summer filter, or is it under OTHER?
6. Related to 5, is there a way to like indicate the course is a study abroad course like our 3390?

My thoughts
1. If a gpa is blank for a Section filter, let's not include it. I think this behavior is fair because there is no indication that that course-prof taught an honors section, for example
2. Now that we're auto-selecting to only show profs teaching next sem, I find this filter a little confusing. Because it will show which profs have historically taught an online section but not necessarily who _is going to be_ teaching an online section. How do you think we can address this @TyHil @Matrixk1ng

AbhiramTadepalli (changes requested):
I'd like the graph bugs (# 1-4) to be resolved before merging pls

AbhiramTadepalli:
@TyHil could you work on this?

egsch:
OK I looked some more and I still do not really understand how to fix bugs 1, 2, and 4 so if anyone else has ideas that would be great! All I really know is that ApexCharts has bad error handling lol but there's gotta be a way to prevent the series from changing and causing the errors

PR #335, opened 2024-11-27
Mobile table usability

RVC13850:

PR #391, opened 2025-03-12
Fix mobile table loading

TyHil:
Missed this in #335 but fixed the loading row for mobile so it looks like this:
image
not this:
image


AbhiramTadepalli (approved):
much needed

PR #392, opened 2025-03-16
Mobile expanded UI

pnagre05:
Objectives
----------------------
Constrain the graph to the width of screen

Squeeze all the statistics together

Font size can decrease slightly

Have the RMP tags wrap better (fit it to the width of screen)


AbhiramTadepalli (changes requested):
so 3 main pieces of feedback 

1. Since you're only trying to change the mobile view, go back in and add breakpoints (I think tyler sent a message about it in #trends-engineering a while back) so that the changes only affect mobile. Not Laptop/PC
2. In SearchResultsTable.tsx  see what you can modify, not remove. ie LoadingRow is an important component. The functionality should not change, just the design
3. The card isn't opening unfortunately. Can you figure out why it's crashing

This was a good attempt, maybe try it from scratch on a nenw  branch using this as reference (that's probably easier than trying to fix this since you removed some stuff)

egsch:
Closing this PR for now since we are not using this branch.

PR #397, opened 2025-03-26
LHS card toggle

TyHil:
Adds a toggle to the LHS cards to show sections or grades.


AbhiramTadepalli (approved):
something about 2 toggles is irking me maybe it's cuz they look the same. But we ball!

PR #415, opened 2025-03-29
Update planner choices across tabs

TyHil:
If you have dashboard open in one tab and planner open in another this will keep them in sync by listening for changes


AbhiramTadepalli (approved):
good job, nice though for QoL

PR #416, opened 2025-03-30
Rule 1 fix for lab+test sections

TyHil:
Fix to allow selection multiple sections but only when they are:
- 1xx: Lab section (sciences)
- 2xx: Discussion section (humanities)
- 3xx: Problem section (maths)
- 5xx: Night lecture (past 5 PM)
- 6xx: Lab night section (past 7 PM)
- 7xx: Exam section

Does this by adding a new SearchQuery type that stores sectionNumbers (an array of strings) instead of sectionNumber (a string)


AbhiramTadepalli (approved):
makes me uncomfy but that's how utd courses are so it's not perfect but on we merge

Issue #369, opened 2025-02-28
[Feature Request] MyPlanner RHS Section Component v2


AbhiramTadepalli:
# Overview

Create the Section Component so that it can be placed in the Schedule and show section information

# Proposal
- [x] Showcase all the necessary information (up to you, but I think course number + section, prof, loc...)
- [x] Handle different widths and heights (play around with the middle slider and maybe open browser tools)
- [x] Support colors that match across LHS and RHS
- [x] Dark mode + Light Mode

## Supplementary Information
Can model it off of out mockup OR Orion schedule planner. Whatever works best. Ik schedule planner conveys all the info it needs to.
Mockup, but it might be a bit outdated

<img width="114" alt="Image" src="https://github.com/user-attachments/assets/19baf948-46fb-4e1c-87d3-af78fd2b6cb9" />

https://www.figma.com/design/OmGLQNdZBrgWGfamGjkci5/Trends-Mockups-24-25?node-id=903-13&t=76SiFzJBo98ltQ1F-0

egsch:
Hey for those of you who are going to be working on this, I have a starting point for the Section component in https://github.com/UTDNebula/utd-trends/tree/367-feature-request-myplanner-rhs-schedule-component that goes with my new Schedule component.

The height of the component doesn't work right now (it shows all classes as 1.25 hours long), so you'll need to fix that, and you'll probably need to change the Schedule and Section component definitions to include props for color coding or something. Feel free to make any changes you need to, and lmk if anything doesn't work or is confusing

AbhiramTadepalli:
> The height of the component doesn't work right now

The height works now. I set it up so in PlannerSection you have both course information as well as the section information. So this is issue is mainly about styling @RVC13850 and @SashaankG 

I've updated the issue tasks above too

egsch:
Just realized I made a mistake in finding the start and end times when I set up the section component because I forgot 12pm exists and added 12 hours to times between 12 and 1, putting them off the edge of the schedule.  ex. classes from 11:30-12:45 show up as 12+ hours long

I don't want to complicate the PRs too much so would y'all mind just changing the places we check for `meeting.start_time.includes('pm')` so it only adds 12 if the time isn't 12pm? If you can't let me know and I can handle it separately.

RVC13850:
i just commited a fix for that

PR #418, opened 2025-03-30
369 feature request myplanner rhs section component v2

RVC13850:
Section data is now fetched dynamically, fixed issues with coloring, schedule size, margins

#369


AbhiramTadepalli (approved):
if it works it works @RVC13850 @SashaankG gj!

PR #419, opened 2025-03-30
Whats new tweaks

TyHil:
- Removed some extra indentation since `Popover` already has a built in `Paper` component accessible in the `slotProps`.
- Changed the buttons to links since we're going to a new page.
- Copied over `usePersistantState` from planner to avoid code duplication later. This stores read feature ids in localStorage instead of the whole feature info


AbhiramTadepalli (approved):
LGTM!

PR #435, opened 2025-04-03
Warning fixes

TyHil:
Fixed most of the ESLint warnings. Still not the ones on `dashboard/index.tsx`.

- `WhatsNew`: moved the fetch logic outside the components so it doesn't need to be in the useEffect dependency array.
- `Carousel`: add children to dependency array and an if statement to only really run when compare length changes.
- `PlannerCard`: use the prev value from setWhichOpen instead of whichOpen.
- `Filters`: use query as source of truth, not states.
- `usePersistantState`: initialValue shouldn't change so save it in a ref and don't update it.
- `planner/index`: useRef to not rerun when `fetchAndStoreGradesData` or the like changes.

### DIFF
--- a/src/components/common/WhatsNew/WhatsNew.tsx
+++ b/src/components/common/WhatsNew/WhatsNew.tsx
@@ -22,92 +22,64 @@ interface Feature {
   releaseId: string;
 }
 
-/**
- * WhatsNewButton component that shows the latest feature from GitHub releases
- */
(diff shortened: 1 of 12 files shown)


AbhiramTadepalli:
will do at the end of my merges

AbhiramTadepalli (approved):
ty

PR #448, opened 2025-04-07
Cutoff issue when prof name is long in myplanner section

TyHil:
Ellipsis instead of text wrap

Also semi related but fixed Inter not being the default font everywhere. It makes everything a bit bigger but I think the font looks nicer


AbhiramTadepalli:
hmm okay so it'll always be 3 lines

AbhiramTadepalli (approved):
oh yeah, sweet sweet inter

PR #449, opened 2025-04-07
Add RMP overwrites

TyHil:
Overwrites fetch for RMP data for Yu Chung Ng to Vincent Ng. Others can be added


AbhiramTadepalli:
Okay now all we need to do is fill the `src/data/professor_to_alias.json` file

AbhiramTadepalli:
https://github.com/UTDNebula/website-v2/commit/8732e319b0aeac875cfeb8b6ed6c30531b3450ca same fix

AbhiramTadepalli:
Okay now all we need to do is fill the `src/data/professor_to_alias.json` file

AbhiramTadepalli:
should we link to the rmp search result page for people w/o profs?

egsch (approved):
OK guys I am super super sorry for letting this sit here so long, it looks really good. If y'all want I can make a separate issue for filling out the failover graph so we can just merge and mark this one done

egsch:
> should we link to the rmp search result page for people w/o rmp?

This is an interesting idea but I'm thinking we probably don't need to? I'll note it down though and we can ask ppl later if they think it would be useful

AbhiramTadepalli:
@TyHil will make it so that it only checks alias as a fallback if there are no matches in RMP

TyHil:
We'll want to do some more testing on this as we add the aliases but graphql is cool

TyHil:
Woohoo, I'll add and test them

egsch (approved):
GraphQL changes look good, great to see it working all the professors!

PR #450, opened 2025-04-07
Fix for courses with no professors

TyHil:
Courses with no professors weren't being added to the combos table which was causing an error. This adds empty arrays for them.


AbhiramTadepalli (approved):
lgtm

AbhiramTadepalli:
@TyHil wait why is there a prof result for this in dev but not in this deployment?
https://dev.trends.utdnebula.com/dashboard?searchTerms=ATCM+3304&availability=true

PR #454, opened 2025-04-19
Optimize section requests

jpahm:
## Overview

This PR seeks to allow Trends to be more performant by optimizing section requests, which take up a large portion of the loading time and visual latency (especially on initial result table loads).

## What Changed

This PR does the following:
- Adds a simple, client-side in-memory section request cache to `useSectionsStore` (helping to prevent multiple overlapping section requests that ultimately fetch the same exact data)
- Modifies the sections API to pull from new Trends-optimized Nebula API endpoints for section data (yes, I cheated and made new endpoints specifically for this)


TyHil:
Merged in app router which should do caching automatically, set to an hour.

PR #456, opened 2025-04-28
Course name search

TyHil:
Made https://github.com/UTDNebula/nebula-api/pull/266 to get the title into the same aggregated data we pull for the normal autocomplete.

## Algorithm

This is a bit brute force but there's only 5,579 courses so it still runs fast. Basically this takes each title and for each word in your search finds the word in the title (or the course number/prefix) with the lowest edit distance, then sums those for a score. So for the search "stte government" and the title "State and Local Government", "stte" would have an edit distance of 1 to "State" and "government" would have an edit distance of 0 to "Government". So the score for "State and Local Government" would be 1+0=1. Then we just return the 10 results with the lowest score. Since its also checking course prefix and number, searching "data structures CS" should place CS 3345 above CE 3345.

## Search Bar

The search bar acts exactly the same when it finds a result for what you type in the normal autocomplete graph. Once the normal autocomplete graph returns an empty array it calls the course name endpoint and displays those results. So typing "da" gives you a bunch of professor names but "dat" gives a bunch of course titles since there're no professors who have a first or last name starting with "dat".


AbhiramTadepalli:
brother wtf

TyHil:
oops

AbhiramTadepalli:
machine learning does not show cs4375. I'll give a shot at rewriting the edit distance algo

TyHil:
seems there are a lot of classes that mention machine learning pushing it down. if you search "machine learning cs" it's in the list

AbhiramTadepalli:
Changes I want to try:
- [ ] only autocomplete results that have every SearchQuery -- so "data structures and introduction to algorithmic analysis" does not get "Introduction to Latin American Studies" at the bottom
- [ ] Use word/query length to sort. So "Machine Learning" shows "Introduction to Machine Learning" higher than "Applied Artificial Intelligence Machine Learning". I can do percentage of result that the user has typed.
- [ ] IF all else fails, rank CS courses higher, then JSOM, etc but I don't like this

TyHil:
Just clarifying, is this a todo list for me or you?

AbhiramTadepalli:
mi

AbhiramTadepalli:
Lots of changes, the commit goes through them all

AbhiramTadepalli:
>     * [ ]  searching "big data" is hard cuz SPACE

And any suggestions for this?

TyHil:
Can you summarize the changes you made by updating the PR description. Also can you remove console.logs and code that got commented out instead of being deleted.

AbhiramTadepalli:
Aero 2040 bugs out on a search, can u figure out why @TyHil

TyHil:
> > ```
> > * [ ]  searching "big data" is hard cuz SPACE
> > ```
> 
> And any suggestions for this?

we could not auto complete a prof name/course unless they've typed the whole first or last name



> Aero 2040 bugs out on a search, can u figure out why @TyHil

merge in develop to get the #450 fix

AbhiramTadepalli (approved):
lgtm

PR #459, opened 2025-05-05
v3.1.0

TyHil:
- Next.js app router server side rendering
- Convenience endpoint for sections
- SEO
- Fix for courses with no professors


TyHil:
Merging so we can stop paying for Vercel hopefully

PR #460, opened 2025-05-25
 Disable optimization on UTD Profiles images 

TyHil:
We're kinda creeping up on our Vercel Image Transformations limit and while Vercel will probably let us go over, I want the whole limit available for a Jupiter launch. All the UTD Profiles images are already only 450x450 so I disabled optimization on them. I think this is the only place Trends really uses images so it should give half the limit back to Jupiter.

image


AbhiramTadepalli (approved):
it's good.

PR #461, opened 2025-05-26
Add Sentry and Bug Report Button

TyHil:
Also removes the old feedback UI so the REACT_APP_GOOGLE_CREDENTIALS env variable can be removed from Vercel and the google cloud console project can be deleted after merge.


AbhiramTadepalli:
image
is that something we can fix or nah?

TyHil:
that just happened to me but it worked when i turned off ublock origin. there's a sentry option to route all sentry communication through a nextjs api route which would fix this entirely but i was hesitant to enable it since we get close to our vercel free limits

AbhiramTadepalli:
leave it as is that's fine

AbhiramTadepalli:
in dark mode it overlaps the github icon, not in light mode

AbhiramTadepalli:
left would maybe overlap a row's compare button

we can remove the github icon. The 'i' at the top does that anyway?

TyHil:
well the github button is for developers and is only on preview deployments to find the commit the preview was made from. the i button is for end users on production releases

AbhiramTadepalli:
ah okay then move it left

AbhiramTadepalli (approved):

PR #462, opened 2025-05-29
Separate styleguidist packages

TyHil:
Finally figured out how to do this. This separates the packages needed for our doc generation into another package.json file. Basically just to help with organization, so when you look at the main package.json file you know that's all there to run Trends with Next. And when you look at the other package.json you know that's only needed for styleguidist.


AbhiramTadepalli (approved):
lgtm

PR #468, opened 2025-06-18
Name autocomplete improvements

AbhiramTadepalli:
A few improvements to https://github.com/UTDNebula/utd-trends/pull/456
- Fix the algo for an edge case involving blank strings after a split (double space ig??)
- Make it so that you have to type the entire name of a prof for 'space' to auto-select their name

### DIFF
--- a/src/components/search/SearchBar/SearchBar.tsx
+++ b/src/components/search/SearchBar/SearchBar.tsx
@@ -198,7 +198,7 @@ export default function SearchBar(props: Props) {
           throw new Error(data.data ?? data.message);
         }
         //remove currently chosen values
-        const filtered = data.data.filter(
+        const filtered: SearchQuery[] = data.data.filter(
           (item: SearchQuery) =>
             !value.some((el) => searchQueryEqual(el, item)),
         );
@@ -213,13 +213,19 @@ export default function SearchBar(props: Props) {
           // so it should autocomplete then when this is realized
           quickInputValue.current.charAt(newInputValue.length) === ' '
         ) {
+          // only add chip on space when it matches the full prof name
           if (
-            ((typeof filtered[0].profFirst === 'undefined' &&
+            (typeof filtered[0].profFirst === 'undefined' &&
               typeof filtered[0].profLast === 'undefined') ||
-              filtered[0].profFirst?.toLowerCase() +
-                ' ' +
-                filtered[0].profLast?.toLowerCase() ===
-                quickInputValue.current.toLowerCase().trim()) // only add chip on space when it matches the full prof name
+            searchQueryEqual(
+              {
+                profFirst: filtered[0].profFirst?.toLowerCase(),
+                profLast: filtered[0].profLast?.toLowerCase(),
+              },
+              decodeSearchQueryLabel(
+                quickInputValue.current.toLowerCase().trim(),
+              ),
+            )
           ) {
             addValue(filtered[0]);
             const rest = quickInputValue.current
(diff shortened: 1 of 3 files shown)


AbhiramTadepalli:
Also I forgot to write the algorithm I made for autocomplete in the last PR, do you still want it? It's commented in the code

AbhiramTadepalli:
`hons 3119` exists in coursebook, not in grade data (first offering is 25F) so it's not in combotable. But since it shows up in coursebook the autocomplete shows it. Selecting leads to error

Do we want to make an empty card for those courses?

AbhiramTadepalli:
unrelated, did we get rid of the surrounding card for RHS in lightmode? I see it in dark mode

TyHil:
ok I thought that would help but i dont get what you said. there's something commented in that code that we might still want to implement? or it needs more comments?

AbhiramTadepalli:
> Can you summarize the changes you made by updating the PR description.
> 
> _Originally posted by @TyHil in https://github.com/UTDNebula/utd-trends/issues/456#issuecomment-2961228408_

do you still want to document it somewhere, confluenc

TyHil:
as long as the comments are updated, fine with me

AbhiramTadepalli:
@egsch we can merge this if it looks good for you

AbhiramTadepalli:
Okay bug related to #471 -- type "short story" rly quickly and it selects kevin short but keeps typing, which is something I aimed to fix. Doing it slowly is fine. @TyHil

TyHil:
> Okay bug related to #471 -- type "short story" rly quickly and it selects kevin short but keeps typing, which is something I aimed to fix. Doing it slowly is fine. @TyHil

There are 2 ways autocompletes on space are made, I think you may have missed changing one. One is when the options are already loaded the the next character is a space, the code checks if there's a match, which I believe you changed. The other is when the fetching is behind and the options are not loaded in which case when the options for "short " comes in it looks for a match, adds the chip, and replaces the current text of "short story" with "story". Should just be another place to put similar if statements.

Also don't forget to remove that console.log

src/components/search/SearchBar/SearchBar.tsx
```diff
@@ -213,12 +213,23 @@ export default function SearchBar(props: Props) {
...
         ) {
-          addValue(filtered[0]);
-          const rest = quickInputValue.current
-            .slice(newInputValue.length)
-            .trimStart();
-          setInputValue(rest);
-          loadNewOptions(rest.trimEnd());
+          if (
+            quickInputValue.current.length > 0 &&
+            filtered.length === 1 &&
```
TyHil, line 218: this and the line above seem redundant based on the outer if statement

src/components/search/SearchBar/SearchBar.tsx
```diff
@@ -213,12 +213,23 @@ export default function SearchBar(props: Props) {
...
+            ((typeof filtered[0].profFirst === 'undefined' &&
+              typeof filtered[0].profLast === 'undefined') ||
+              filtered[0].profFirst?.toLowerCase() +
+                ' ' +
+                filtered[0].profLast?.toLowerCase() ===
```
TyHil, line 223: is there any reason not to use searchQueryLabel here?

TyHil:
done. i had been trying to add my own keyboard handler for that before which MUI annoyingly disables. turns out this method lets you click out to close the tutorial too!

also left a couple questions

src/components/search/SearchBar/SearchBar.tsx
AbhiramTadepalli: case-sensitive

AbhiramTadepalli:
@egsch good to review now

egsch:
image
These changes look good! I did happen to notice that filtering out blank splits brought back this weird behavior where it suggests blank titles after you typing a short non-prefix + a space. I think that the filtering change should still be made, it's not causing this directly and this goes away as soon as you type more but just wanted to check if y'all think it's worth trying to fix

egsch (approved):
super sorry for taking so long but this looks amazing!

Issue #267, opened 2024-10-19
[Feature Request] Search Engine Optimization (SEO)


AbhiramTadepalli:
# Overview

Describe the motivation behind your desired functionality.

How would you like the user experience for this feature to work?

# Proposal

There are a couple parts to this. First we need to find out what meta data we can customize for our site that SE's read. Then we need to identify good keywords to use when filling those out. Also sitemap. After that, let's try to get Trends to be #1 for different search queries (like UTD Galaxy, Orion, Coursebook, etc)

## Tasks

- [ ] What metadata can we customize for search engines? Comment it below
- [ ] Important keywords
- [ ] Fill out the meta tags
- [x] Finalize a sitemap (should be very simple imo)
- [ ] Incorporate keywords to put Trends at the top of tangentially-related search results

AbhiramTadepalli:
https://www.google.com/search?q=utd+trends+site:utdnebula.com&sca_esv=2100fb941db67f3a&sxsrf=ADLYWIL3MU0wPovHPU1KMY6rrLjanRnZ5g:1730841586874&sa=X&ved=2ahUKEwiVxsqvj8aJAxXjLdAFHdUJFAIQrAN6BAgMEAE&biw=948&bih=724&dpr=1.25

for reference

PR #476, opened 2025-07-04
Big sitemap

TyHil:
This programmatically includes every possible search term in the sitemap with a handy Next.js feature: https://nextjs.org/docs/app/api-reference/file-conventions/metadata/sitemap. So ideally if it all gets indexed by Google, any course or professor you look up could have a trends link in the results. It's about 7.5k entries which is under the 50k limit for a single sitemap. It's also 1.55 MB uncompressed.

See https://utd-trends-git-big-sitemap-utdnebula.vercel.app/sitemap.xml.

For #267, not sure if this should resolve it tho.

### DIFF
--- a/src/app/dashboard/page.tsx
+++ b/src/app/dashboard/page.tsx
@@ -9,6 +9,7 @@ import Filters, { LoadingFilters } from '@/components/search/Filters/Filters';
 import { LoadingSearchResultsTable } from '@/components/search/SearchResultsTable/SearchResultsTable';
 import {
   decodeSearchQueryLabel,
+  type SearchQuery,
   searchQueryLabel,
   searchQuerySort,
 } from '@/types/SearchQuery';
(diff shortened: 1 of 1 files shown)


AbhiramTadepalli:
can/should we include "CS 1200 John Cole"?

egsch (approved):
This looks great! Very exciting!

AbhiramTadepalli (approved):
looks good! Just wanted to check sometimes google punishes a website if it has like a bunch of links all at once, this is fine?

TyHil:
did some research and it seems fine as long as there aren't any "broken links, redirects, or pages with noindex tags". i tested some random links and didnt find any broken ones. i think we should be good.

that being said we have had a few issues in the past with broken pages like with ATCM 3304. that's all fixed but other stuff could be lurking. we'll just have to fix it as it comes up

AbhiramTadepalli:
sounds good

PR #478, opened 2025-07-17
Account for extra URL encoding

TyHil:
In a URL " " is replaced with "+" and in URL encoding "+" is replaced with "%2B" which we were not accounting for:

http://dev.trends.utdnebula.com/dashboard?searchTerms=MATH%2B2413
https://utd-trends-git-url-encoding-fix-utdnebula.vercel.app/dashboard?searchTerms=MATH%2B2413

This PR uses `decodeURIComponent` to convert convert an extra percent encoding back to "+", ",", etc.

### DIFF
--- a/src/app/dashboard/page.tsx
+++ b/src/app/dashboard/page.tsx
@@ -28,7 +28,7 @@ export async function generateMetadata({
     searchTerms = searchTerms[0];
   }
   if (typeof searchTerms !== 'undefined') {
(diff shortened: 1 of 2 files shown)


AbhiramTadepalli:
Good catch @egsch, this should fix it

AbhiramTadepalli (approved):

egsch (approved):

PR #479, opened 2025-07-17
add extra lab/test sections in a card

AbhiramTadepalli:


AbhiramTadepalli:
Thank you Mai!!

egsch:
Note: 8XX classes are currently counted as lecture sections, but seem to actually be evening math problem sections

Issue #471, opened 2025-06-26
[Bug]: Quickly pressing enter on search does nothing and removes query


AbhiramTadepalli:
### What did you expect to happen?

When I type "CS 1200" and press enter quickly, before the autocomplete loads, my search query gets deleted.

### What happened?

Either:
(a) Upon clicking enter, wait until the autocomplete loads, then register the enter action
(b) Upon clicking enter, do nothing if autocomplete has not loaded.

### Version

trends.utdnebula.com

### What browsers are you seeing the problem on?

Firefox

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #504, opened 2025-09-14
Issue #471

sriyuthy:
## Overview

Fixes issue #471

Added conditionals for when user clicks enter in search bar: checks if user input value contained anything and if there were any results in autocomplete, and based on that either searches with selected values or extracts SearchQuery object from what user entered and uses that value to search

## What Changed

Modified SearchBar.tsx

### DIFF
--- a/next-env.d.ts
+++ b/next-env.d.ts
@@ -1,6 +1,5 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-/// <reference path="./.next/types/routes.d.ts" />
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/app/api-reference/config/typescript for more information.


src/components/search/SearchBar/SearchBar.tsx
```diff
@@ -140,10 +140,25 @@ export default function SearchBar(props: Props) {
...
+      if (inputValue === "") {
+        onSelect(value);
+      }
+      else if (options.length > 0) {
+        addValue(options[0]);
+        setOptions([]);
+        setInputValue("")
+        onSelect([...value, options[0]])
+      } else if (inputValue !== "") {
+        const searchQuery: SearchQuery = decodeSearchQueryLabel(inputValue)
```
egsch, line 155: Unfortunately I think this might need to change because our goal in searching is that users can only search for complete classes or professors. With this change, if you type in something like "cs 123" and then press enter really fast, it will try to search for "cs 123" in the search results and then show an error when that class doesn't exist.

 If we can wait for the new options to load first, and then select the first one (like you do above, but with the updated suggestions), that would be great. If not, we can keep what we've typed and avoid doing anything else on enter if the options are empty, but we should make sure users can't search for incomplete class or professor names

egsch (commented):
Hey thank you for working on this! Sorry, I think I underestimated how much unclear code there is to wade through here - I added a couple comments about some potential issues, please read them and ask any follow up questions

src/components/search/SearchBar/SearchBar.tsx
```diff
@@ -140,10 +140,17 @@ export default function SearchBar(props: Props) {
   }
 
   function handleKeyDown(event: React.KeyboardEvent<HTMLInputElement>) {
-    if (event.key === 'Enter' && inputValue === '') {
+    if (event.key === 'Enter') {
       event.preventDefault();
```
egsch, line 144: Sorry for yet another question - if preventDefault and stopPropagation apply to all 'Enter' events, do they prevent selecting by pressing enter when there are autocomplete options? 

If so, could we change this so these only occur when loading or when inputValue is empty?

next-env.d.ts
```diff
@@ -1,5 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
+/// <reference path="./.next/types/routes.d.ts" />
```
egsch, line 3: When you rerun `npm run dev` with your current branch code, does this change back to what it was before or does it stay the same?  It seems like this shouldn't be changing with your current code

PR #505, opened 2025-09-15
Issue 483: Adding Undefined GPA to Compare Crashes Out

Evelas78:
Implemented quick conditional to prevent fully undefined gpa from crashing website. For example, all withdraws in 2V90.


egsch (commented):
Thanks for adding this, this is a good backup for classes with only non-numeric grade data as we're working on incorporating this data better. 

One thing, we have a lint and formatting check that automatically runs on PRs, and it looks like it found a formatting issue - you can auto-fix them by running `npm run format`

Issue #354, opened 2025-02-18
[Feature Request] Recent Searches


AbhiramTadepalli:
# Overview
Show the 3 most recent SearchQueries in autocomplete

# Proposal
- [ ] Store the 3 most recent SearchQueries (that are not currently in the SearchBar)
- [ ] Display these when there is nothing being typed but the searchbar is selected as autocomplete options
- [ ] Move _matching_ recent searches to the top of the autocomplete list when the user has typed text

## Supplementary Information
Design is to-be finalized, but the logic can start being worked on

KJ14GOD:
will try to do this

egsch:
Assigned to @Shurgbee (assigning myself as a placeholder pre-Github access)

TyHil:
Hey yeah local storage is only really used by my planner now so no issues with it filling up! Would be great to keep recent searches in there

PR #506, opened 2025-09-21
Recent searches

shurgbee:
## Overview

Addresses #354 

## What Changed

The last 3 searches in the searchbar are stored in localstorage as UTDTrendsRecent. This gets displayed in the autocomplete with either a search or history icon next to it.

## Other Notes

Slightly modified search query type


AbhiramTadepalli:
Hey @shurgbee great work! Before we get this to prod, I wanted to do some rigorous testing and came across these things:
1) If I've searched for GOVT 2306 in the past, could we make it so that typing "gov" shows that recent query in the list?
2) Once I search for CS 1200 and GOVT 2306, I can alternate between adding them ad infinitum
image

shurgbee:
I added the substring search functionality and I think it should be ready to ship.

egsch (approved):

AbhiramTadepalli:
@shurgbee @egsch I caught an edge case added by the slight complexity of maintaining an auxiliary recentSearches array, so I'm refactoring your code to work with just the filtered array :) Finishing touches now

egsch (approved):

Issue #437, opened 2025-04-06
[Feature Request] Export/Share schedule


AbhiramTadepalli:
# Overview

Allow exporting the schedule as a png/image

## Supplementary Information

Research what library can allow you to basically take a screenshot. You might have to remake the schedule as an html page

PR #508, opened 2025-09-22
[Feature Request] Export/Share schedule #437

PyroSh0ck:
## Overview
Resolves Issue #437 - Export/Share Schedule

This PR adds the functionality to download the planner as an image, with minimal modifications to other components and allowing for scalability with future UI changes to the planner, or the desired downloaded image.

## What Changed
Imported the html2canvas-pro library, which is used for creating a canvas image based on the DOM for a component, with the pro version allowing html2canvas to work with previously unsupported CSS attributes and properties. Modified the TopMenu component to show a "download" favicon, instead of the "share" favicon, from MUI if isPlanner prop is true. Likewise, changed the tooltip to "Download your Schedule". Then added the onClick functionality for the button, where the html2canvas function is used to convert the referenced component to a canvas image, which is then promptly turned into a blob. This blob is given an objectURL, which is set as the href for a newly-created "a" tag. The "a" tag is added to the DOM, clicked via the script to download the image, then removed from the DOM. The component is referenced through a useRef placed in the PlannerPage component (a new client component wrapping the TopMenu, Planner, and PlannerDownloadUI components). The useRef is passed as a prop to PlannerDownloadUI, which is another new component. The aim of this component is to be able to easily modify the downloaded image, without changing the base page. It contains the PlannerSchedule component, which allows it to be updated with any changes to the planner from the client, as well as any ui design changes. Lastly, the PlannerDownloadUI component is moved very far off screen using "left-[-9999px]", to ensure that it stays in the DOM (hence allowing html2canvas to turn it into a canvas image), while not being visible to the user. The original page.tsx file now only contains the PlannerPage component.

## Other Notes
An alternative to the PlannerPage component may be desired.


src/app/planner/PlannerDownloadUI.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import NebulaLogo from '@/components/icons/NebulaLogo/NebulaLogo';
+import PlannerSchedule from '@/components/planner/PlannerSchedule/PlannerSchedule';
+
+
+
+export default function PlannerDownloadUI({ downloadRef } : { downloadRef : React.RefObject<HTMLDivElement | null> }) {
+  
+    return (    
+        <div className='bg-black p-4 absolute left-[-9999px]' ref={downloadRef}>
```
egsch, line 13: Can we add `top-0` to this so it doesn't impact the page length and sticky scrolling while not visible?

src/components/navigation/TopMenu/TopMenu.tsx
```diff
@@ -21,6 +23,7 @@ import SearchBar, {
  */
 type Props = {
   isPlanner: boolean;
+  downloadRef: React.RefObject<HTMLDivElement | null>;
```
egsch, line 26: I know you're aware of this already but would you mind doing the changes to make this optional?

src/app/planner/PlannerDownloadUI.tsx
```diff
@@ -0,0 +1,24 @@
...
+            <div className='flex flex-row mb-4 justify-start'>
+                <NebulaLogo className="h-6 w-auto fill-haiti dark:fill-white mr-2 ml-2" />
```
egsch, line 15: Another really small design thing, I noticed the left edge of the nebula logo is getting clipped off on my computer. Could you try moving the margins to prevent this?

egsch (commented):
I am so so sorry for the late response, thank you so much for your work on this - it looks good and works really well! Good work finding and integrating the html2canvas-pro and the download looks great! Added a couple small comments, again sorry to make these so late

PR #511, opened 2025-09-30
Added filtering functionality to line graph

Evelas78:
Used ApexChart's built in markerClick function.


AbhiramTadepalli (approved):
looks great!

src/components/graph/LineGraph/LineGraph.tsx
```diff
@@ -68,6 +69,7 @@ type Props = {
 };
 
 export default function LineGraph(props: Props) {
+  const { semesters, chosenSemesters, setChosenSemesters } = useSharedState();
```
egsch: will need to update this reference to `semesters` after state refactor

PR #514, opened 2025-10-01
486 bug] line graph and mean are not matching up

ridham15:
Problem: Summer semester was not included in the line graph. 

Added summer points to the line graph along with the regular semester on the same line and removed the summer labels from the xaxis. This created empty lables where the summer points are and removing this messes up the alignment of other points. 
Screenshot 2025-09-29 152050

### DIFF
--- a/src/components/graph/LineGraph/LineGraph.tsx
+++ b/src/components/graph/LineGraph/LineGraph.tsx
@@ -27,10 +27,9 @@ function getSemesterGPAs(
     name: string;
     data: Grades['grades'];
   },
-  semesterIndex: Map<string, number>,
+  semesterMapping: Map<string, number>,
 ) {
-  const regular: { x: number; y: number }[] = [];
-  const summer: { x: number; y: number }[] = [];
+  const allPoints: { x: number; y: number }[] = [];
 
   data.data
     .toSorted((a, b) => sortSemesters(a._id, b._id))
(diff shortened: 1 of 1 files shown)


ridham15:
> Hey this looks good, great job on getting the points to align and making it readable in mobile! I did notice one tiny thing which was that, for search results that were only taught in one semester, the points are no longer centered in the graph and are instead pushed to the left side. Do you think you could keep them centered?

ridham15:
> > Hey this looks good, great job on getting the points to align and making it readable in mobile! I did notice one tiny thing which was that, for search results that were only taught in one semester, the points are no longer centered in the graph and are instead pushed to the left side. Do you think you could keep them centered?

ridham15:
Do you have the class or professor you saw this issue with so I can test it out?

AbhiramTadepalli:
> dyk why sometimes it shows 25F in the x axis, sometimes it doesn't? Look at Douglas Dow vs Euel Elliot for GOVT 2306

@ridham15

ridham15:
> dyk why sometimes it shows 25F in the x axis, sometimes it doesn't? Look at Douglas Dow vs Euel Elliot for GOVT 2306
I am working on it rn. But I think its because its not checking which labels have data and the labels are aligned with X-axis values where Spring/Fall semesters use integers (0, 1, 2…) and Summer semesters use fractional values (0.5, 1.5…), so the axis could extend beyond the actual data and show extra labels

egsch:
> @egsch I made a new version with a extra label on either end. Is this fine or should I make them empty labels?

The padding labels are a good idea but I think it's better to avoid them because they take up space in already-crowded graphs. Also looks like there are still some issues with summer classes with the padding:
image

ridham15:
> The padding labels are a good idea but I think it's better to avoid them because they take up space in already-crowded graphs. 

the empty labels will also take up the same space so I thought its better to have text.

egsch:
> > The padding labels are a good idea but I think it's better to avoid them because they take up space in already-crowded graphs.
> 
> the empty labels will also take up the same space so I thought its better to have text.

Ok never mind, I think I may have misunderstood your original comment. Having text on padding labels is fine, my note on taking up space was in response to classes starting/ending with fall semesters having padding on both sides, this might be a bug:
image

From what I saw pre-this last commit it seemed good but I may have missed some cases where an initial summer grade got cut off

PR #516, opened 2025-10-07
Feature/calendar preview

davis118:
# Calendar Preview Feature

## Overview

Adds calendar preview system allowing users to hover over course cards to see sections overlaid on the schedule before adding them to their planner.

## What Changed

**Preview System:**

- `PlannerCard` now updates `previewCourses` state on hover/open
- `PlannerSchedule` displays preview sections, filtered to exclude conflicts with existing planner courses
- `PlannerSection` gets `onSectionClick` prop for selecting preview sections
- Previews shown in greyscale, hover adds back color
- `PreviewSectionGroup` handles multiple sections at same time with popover

**Scoot Algorithm:**

- Courses that overlap "scoot" to not overlap

**UI Enhancements:**

- Red border on hover for visual feedback

**State Management:**

- `getSelectedSections` and `hasConflict` moved from `PlannerCoursesTable` to `SharedStateProvider.tsx` for reuse
- Preview courses stored as `SearchQueryMultiSection[]` in shared state
- Preview courses clear correctly with browser back/forth navigation


egsch:
>  I am not able to un-click on the RHS to remove a selected class

What would y'all think of scrolling to the selected class's card on click instead of just automatically removing it?

> > I could also easily filter out the previews that conflicted and did that.
>
> I just wanted y'all's opinion on this. Some configuration could exist where when I open a card, every preview section has a conflict so nothing changes in the RHS. Should we perhaps still do at least a colored outline over that time range?
> image
> 
>  How should we handle the case when multiple cards are open and 2 preview sections overlap?

Ideas for displaying multiple sections at the same time (bad mockups, I am not a designer, please make them look better) after a little bit of discussion at our meeting yesterday:

If classes exactly overlap, put indicator that there are multiple and some kind of popup to choose which one:
image
If overlap with existing thing, same thing but box keeps info/shading:  
image
If partially overlap, keep two separate boxes but raise whichever one starts later to the top and scoot over its left side slightly to indicate that there are multiple sections.
image

Do y'all have any thoughts on these?

AbhiramTadepalli:
Merge #534

AbhiramTadepalli:
Can probably make the bg color of the multiple white in light mode

AbhiramTadepalli:
Need to test on sections with multiple profs

AbhiramTadepalli:
> Add Cs 2336 and govt 2306's overall results and open both cards. In the above image, HN4 actually belongs to the GOVT class
> 
> HOWEVER, @egsch do you think my issue can fix that? Maybe we can leave the grouping title to "Overlap CS 2336, GOVT 2306")

fixed

egsch:
Sorry for the delay. Just updated to handle all sections of selected and non-selected sections together to prevent overlap issues.

>     - [ ]  Need to fix this
> 
>     * [ ]  Handle multiple profs
> 
>     * [ ]  Light mode multiple sections

Will finish these up soon

> 
>     * [ ]  In these cases, is this behavior correct?
> 

I think 1 and 2 are in line with our current expectations but if you have better ideas we can definitely consider them! I think we want to handle scoot in classes that overlap with already-scooted classes a little differently and maybe process different days separately to deal with 3 and 4 (and the issues in the checkboxes above). I will work on these a little more and push something for them soon

mikehquan19:
@egsch Hey I know this one has a whole lot of merge conflicts, can I help yall with resolving it and also any other problems?

PR #518, opened 2025-10-14
State refactor

nl32:
## Overview

Rework state management for more straight forward data flow and simplified state. also gets rid of `syncServerDataToContext`. Implements client side fetching for planner
### Dashboard
- Instead of searchResults being fetched and uplifted into client state through `syncServerDataToContext`, search results are bundled together into a SearchResults object and the list of those is passed down from the server into `ClientLeft`, `Filters`, `Right`.
- `syncServerDataToContext`  Is Removed
- Changes to data access and type definitions through the dashboard to fit with this new shape
- Search result fetching is done in `page` instead of `ServerLeft` so that `filters` and `Right` can access the data.
### Planner
- Data for classes added to planner is fetched client side through a new api route
  - This was being done before using a server action, which is a blocking action and not it's intended use case. now it use React Query in order to fetch the data, when courses are added from the dashboard the data will be added to the query cache to avoid extra fetches
- `syncServerDataToContext` is removed

### Overall
- SharedStateProvider had lot's of state removed as it doesn't handle fetched data anymore
- fetches were redesigned to mostly just throw errors instead of using GenericFetchedData, this means error handling for fetches has to be done at the top level but we already had minimal 
- grades, rmp, section data has been grouped into `SearchResult` objects containing the related information for each course/professor/combo. 

## Unresolved issues

- [x] The Semester filter behavior is not in line with current live behavior
- [x] hydration issues for features reliant on local storage
- [x] Grade Colors changing after load on dark mode due to the use of the `useMediaQuery` hook

## Other Notes

Probably a bit excessive to do a line by line review on this one, the most important files to take a look at will be `SharedStateProvider`, all the `fetch*` files, and Planner files.

### DIFF
--- a/src/components/search/Filters/Filters.tsx
+++ b/src/components/search/Filters/Filters.tsx
@@ -115,29 +115,6 @@ export default function Filters({
   }
   const filterNextSem = searchParams.get('availability') === 'true';
 
-  const filteredResults = useMemo(
-    () =>
-      searchResults.filter((result) => {
-        if (
-          typeof minGPA === 'string' &&
-          calculateGrades(result.grades, chosenSemesters).gpa <
-            parseFloat(minGPA)
-        )
-          return false;
-
-        // check if this search result should have RMP data
-        if (result.type !== 'course') {
-          if (
-            typeof minRating === 'string' &&
-            result.RMP &&
-            result.RMP.avgRating < parseFloat(minRating)
-          )
-            return false;
-        }
-        return true;
-      }),
-    [searchResults, minGPA, minRating, chosenSemesters],
-  );
   function getRecentSemesters() {
     // get current month and year
     const today = new Date();
@@ -173,15 +150,28 @@ export default function Filters({
 
   minGPAs.forEach((gpaString) => {
     const gpaNum = parseFloat(gpaString);
-    gradeCounts[gpaString] = filteredResults.filter((value) => {
-      const courseGrades = value.grades;
+    gradeCounts[gpaString] = searchResults.filter((result) => {
+      if (result.type !== 'course') {
+        if (
+          typeof minRating === 'string' &&
+          result.RMP &&
+          result.RMP.avgRating < parseFloat(minRating)
+        )
+          return false;
+      }
+      const courseGrades = result.grades;
       return courseGrades && calculateGrades(courseGrades).gpa >= gpaNum;
     }).length;
   });
 
   minRatings.forEach((ratingString) => {
     const ratingNum = parseFloat(ratingString);
-    rmpCounts[ratingString] = filteredResults.filter((result) => {
+    rmpCounts[ratingString] = searchResults.filter((result) => {
+      if (
+        typeof minGPA === 'string' &&
+        calculateGrades(result.grades, chosenSemesters).gpa < parseFloat(minGPA)
+      )
+        return false;
       return (
         result.type !== 'course' &&
         result.RMP &&


src/components/DocsWrapper.tsx
```diff
@@ -11,7 +11,9 @@ export default function DocsWrapper({ children }: React.PropsWithChildren) {
   return (
     <div className="bg-white dark:bg-black text-haiti dark:text-white">
       <ThemeProvider theme={theme}>
-        <SharedStateProvider>{children}</SharedStateProvider>
+        <SharedStateProvider latestSemester="Fall 2025">
```
AbhiramTadepalli: Why is this hardcoded?

If it's important, we could totally use like `getCurrentSemester()` or something of that nature

src/components/search/SearchResultsTable/SearchResultsTable.tsx
```diff
@@ -468,12 +467,6 @@ export default function SearchResultsTable({
     }
   }
 
-  // To avoid hydration errors
-  const hasHydrated = useHasHydrated();
```
AbhiramTadepalli: so because this was removed we have hydration errors now?

AbhiramTadepalli (changes requested):
- MyPlanner
- Semesters
- Hydration

nl32:
> > ```
> > * [ ]  Grade Colors changing after load on dark mode due to the use of the `useMediaQuery` hook
> > ```
> 
> How can I replicate this?

set yourself to dark mode and refresh page, they'll load in with the light mode colors and then change to the dark mode ones.

src/components/DocsWrapper.tsx
nl32: `fetchCurrentSemester()` is async and I don't think that styleguidist has support for RSC/async components, also the styleguidist config hasn't been touched with 2022 and I didn't even know it was a thing so I figured hardcoding it for now was easier than figuring out this issue in this pr.

src/components/planner/PlannerCoursesTable/PlannerCoursesTable.tsx
```diff
@@ -53,69 +49,40 @@ export default function PlannerCoursesTable() {
     }
     setOpenConflictMessage(false);
   };
-
+  const allResults = useSearchresults(planner);
+  const latestSections = allResults.map((r) =>
```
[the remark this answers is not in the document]
nl32: you're right this should already be happening in the `api/planner` route, although  the filtering was broken (fixed now), I'll remove these duplicate checks.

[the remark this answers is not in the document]
egsch: > you're right this should already be happening in the `api/planner` route, although the filtering was broken (fixed now), I'll remove these duplicate checks.

just fyi in the most recent deployment it looks like the filtering is still not working the first time a result is added to planner

src/app/Home.tsx
```diff
@@ -133,12 +131,10 @@ export default function Home(props: Props) {
...
               label={
-                'Teaching ' +
-                (typeof props.latestSemester !== 'undefined' &&
-                props.latestSemester.message === 'success'
-                  ? 'in ' +
-                    displaySemesterName(props.latestSemester.data, false)
-                  : 'Next Semester')
+                props.latestSemester == ''
+                  ? 'Teaching Next Semester'
+                  : 'Teaching in' +
```
egsch: Can you add a space here?
image

egsch (approved):
This looks really good and I'm not seeing any issues, thank you so much for all the work you put into it and this rush today!

AbhiramTadepalli:
wow wait nvm I was somehow on the wrong deployment all good

This fixes the issue in Trends with SE 4351 not showing Pushpa Kumar rn

Issue #289, opened 2024-10-29
[Feature Request] Indicate how many results for each filter option


AbhiramTadepalli:
# Overview
Saves time for users if they can see if certain filters lead to 0 results
# Proposal

Include the count of results next to each filters option, like A (58) B+ (47) B (40)...
## Supplementary Information

Might require preprocessing filter partitions or counting on api-load

AbhiramTadepalli:
@justjaiom were you working on this in a branch?

Kutsal04:
Just assigned, looking forward to doing this!

egsch:
Assigned to @Axv2655 (assigning myself as a placeholder until they get Github access)

PR #520, opened 2025-10-15
Completed majority of the feature request

axv2655:
## Overview

Did feature request #289 

rest of the project.

Formats the data and uses a count for each filter value

Only real thing that was modified filters.tsx and I used the same logic to format the query from Clientleft.tsx so I dont need to pass it in as a prop and do it on the master component. 

If were roadblocks encountered during development that remain unresolved or any
future additions or changes to make, note them here. Otherwise. Feel free to
delete this section if it isn't needed.
Wasn't able to integrate the semesters with this, it was extremely difficult to work with and I wasn't able to apply the filter count to it


AbhiramTadepalli (approved):
Great work @axv2655 !!

Issue #470, opened 2025-06-26
[Bug]: In light mode, dashboard LHS and RHS do not have a background


AbhiramTadepalli:
### What did you expect to happen?

Light mode should have a background behind the RHS and the LHS sections like it is in dark mode:

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/682c4d66-0a1d-4cd6-8576-965ce33132e0" />

### What happened?

There is no equivalent background in Light mode: 

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/f186f5f6-c9e4-49ed-8961-c0a6de8ae9fa" />

I would like to see a slight gray bg.

### Version

dev.trends.utdnebula.com

### What browsers are you seeing the problem on?

Firefox

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

egc0205:
Sorry been busy, I have the background changed. I just need to get the table header fixed and I'll be done.

egc0205:
I think the branch should be published now.

PR #526, opened 2025-10-19
Issue#470 lhs and rhs no background

egc0205:
## Overview
Resolves Issue #470 LHS and RHS no background

In layout.tsx I changed the light mode background to be a light gray (246,246,246). In SearchResultsTable.tsx I made a function called getCellSx to make the table head background to be a light gray as well (252,252,252) and I applied this function to the necessary cells for the table head.


AbhiramTadepalli (approved):
looks great!

PR #532, opened 2025-10-22
Standardize colors

TyHil:
All purples should be using royal on light mode and cornflower-300 on dark mode now.

Also made search button match the height of the search bar.

I'd include screenshots but doing light/dark before/after for 3 pages is a lot.


egsch (approved):
@TyHil This looks good, just if we could add rounded corners to the My Planner button like Abhiram said that would be great

Also on myplanner--especially light mode--having the same purple on both sides of the schedule does stand out quite a bit and throw the page design off a little, would it defeat the point to add another intermediate purple color there?

egsch:
> Got it. I'll round the my planner button
> 
> I'm kind of a fan of the single color on the schedule but I have been looking at it too much. I can definitely switch it back to 2 colors tho, up to you.
> image

OK yeah we can keep the one-color thing. I'll merge

Issue #530, opened 2025-10-22
[Bug]: MyPlanner: Switching Between Two Conflicting Sections of Course Fails


egsch:
### What did you expect to happen?

When switching between two sections of the same class, the new section should replace the previously selected section. Currently, if the two section times conflict, you cannot switch between them and our "This section conflicts with your schedule!" popup appears
- Make sure this conflict detection does not prevent switching classes

### What happened?

"This section conflicts with your schedule!" and selection is disabled. Ex. Math 2418 .002 and .004

### Version

trends.utdnebula.com

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Issue #438, opened 2025-04-06
[Bug] Handle professor sections with No meeting time/room (don't show on schedule, but select in LHS)


AbhiramTadepalli:
# Overview

FILM 1303 has sections without a meeting time or room. These are select-able but are not indicated well in the LHS

AbhiramTadepalli:
don't show on schedule in RHS, but select in LHS. I'll put it as just the section number ig?

PR #534, opened 2025-10-27
Extra Sections in MyPlanner + Refactor to take advantage of /sections/trends api endpoint fully

AbhiramTadepalli:
## Overview

Replaces https://github.com/UTDNebula/utd-trends/pull/479

1) Based off of @nl32's refactor, I utilize the `course_details` and `professor_details` introduced in https://github.com/UTDNebula/nebula-api/issues/281 for the `/course/sections/trends/` endpoint to reduce the number of API calls to 1 per unique course + 1 per unique professor.
2) Separate out extra sections like labs and exams into sub-cards in each LHS MyPlanner card. Also add a subcard for unassigned sections (sections of the course without a professor).

## What Changed
**Functionality should be the same as ethan's refactor, plus:**
- Selecting a section should reflect in both the combo card and the overall card
     - Selecting/Deselecting a section should work from both overall and the combo card. 
     - It will add that prof (or overall, for an unassigned section) to `planner` and MyPlanner **if** it is not already added, then it will show up on that added-card's meeting chip I think
- Adding from an overall course should now show the prof name in the RHS card
- syllabus should be how it was before ethan's refactor (chooses most recent of prof, else chooses most recent of the course
    - except, for labs & exams, chooses the most recent syllabus from those sections (`/^[12367]/.test(section);`)
    - and for unassigned professors, keep no syllabus
- Can always switch between sections of a course, even when there's a conflict with an existing selected section of the same course 
     - fixes https://github.com/UTDNebula/utd-trends/issues/530
**The sub-cards**
- Split all the sections of a course into
     - all the sections of that course and professor (or just all the sections for overall) that are NOT extra sections go in the card itself
     - All unassigned profs go in a sub-card (in combos cards only)
     - All Extra sections (`/^[1236]/.test(section)`, not exams) go in a sub-card
     - All Exam sections (`/^[7]/.test(section)`) go in a sub-card
## Other Notes
Still want to do 
- [x] (1) If I select a section in one card,  make sure the correct meeting chips are shown across
- [x] (2) how to show meeting chips for sub-section cards? -- +1 in the overall, then show in the subcard. 
- [x] (2.5) Should only choose 1 lab right?
- [x] (3) how to show meeting chips when selecting extra sections in a combo card and the overall card
- [ ] (4) Show overflow in RHS on hover ig -- (have a "condensed" state to like limit numprofs to the size)
- [x] (5) **Styling the sub-cards**
- [x] (6) Multiple Professors
- [ ] (7) make dashboard's sections call more optimized
- [x] (8) https://github.com/UTDNebula/utd-trends/issues/438
- [ ] (9) Check hydration errors

### DIFF
--- a/src/modules/fetchGrades.ts
+++ b/src/modules/fetchGrades.ts
@@ -58,8 +58,6 @@ export function calculateGrades(grades: GradesData, semesters?: string[]) {
     median_gpa = GPALookup[medianIndex];
   }
 
-  // console.log("bb", grade_distribution)
-
   return {
     mean_gpa: mean_gpa,
     gpa: median_gpa,
--- a/src/app/api/planner/route.ts
+++ b/src/app/api/planner/route.ts
@@ -35,13 +35,6 @@ export async function GET(request: Request) {
       profFirst: profFirst,
       profLast: profLast,
     });
-    // const latest = await fetchLatestSemester();
-    // const filteredResult = {
-    //   ...result,
-    //   sections: result.sections.filter(
-    //     (s) => s.academic_session.name === latest,
-    //   ),
-    // };
 
     return NextResponse.json(
       {


src/modules/fetchGrades.ts
```diff
@@ -58,6 +58,8 @@ export function calculateGrades(grades: GradesData, semesters?: string[]) {
     median_gpa = GPALookup[medianIndex];
   }
 
+  // console.log("bb", grade_distribution)
```
nl32: remove commented out code

PR #535, opened 2025-10-28
fix: Search Filter Counts

egsch:
## Overview

Current Filter counts on 'rating' filter don't account for semesters filter, and founts on both filters don't include classes without grades/ratings, leading to some counts that look incorrect

## What Changed

Count only results that haven't been filtered out by the semesters filter, and count results that have no grade/rating data and are therefore not filtered out


AbhiramTadepalli (approved):
check how the teaching next sem filter works (or tell me what I can look for in testing)

otherwise LGTM

Issue #545, opened 2025-11-12
[Bug]: Compare Chart Error Handling


egsch:
### What did you expect to happen?

When the semester filter is modified, courses that are left with no remaining grade data should be handled gracefully in the Compare Chart.

### What happened?

Empty series cause unhandled errors in our compare bar graph. For example, add a course combo to Compare, deselect semesters until it has no data, and then add another course that still has grade data. An error will occur and the chart will not properly display the new data. 

To resolve this, look at how our bar graph handles empty series and try to prevent ApexCharts from trying to render them.

### Version

dev.trends.utdnebula.com

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #552, opened 2025-12-03
545 graph bug

OddEyed45:
## Overview

List the GitHub issues containing the issues relevant to this pull request.
Resolves issue #545 

rest of the project.

It fixed a simple error where the bar chart does not show up properly on the screen under Compare. 

## What Changed

The issue is fixed by correcting the grade data that is being displayed by the bar graph (under src/components/compare/Compare.tsx). The problem was that if the size of the grade data were 0, the convertNumbersToPercents function would output a list with NaNs since it would not go into the map function. The most direct way to fix this was to check the output of this method (and manually assign an array of 14 zeroes if NaN is found).

## Other Notes
At one point, Vercel failed to deploy because the project was using a version of Next.js with a security vulnerability. I have fixed it on my end (hence, Vercel is working) but I am unsure about changing package.json & package-lock.json for the entire project.

Please make sure to fix that (there is a merge conflict here).

### DIFF
--- a/src/components/compare/Compare/Compare.tsx
+++ b/src/components/compare/Compare/Compare.tsx
@@ -77,11 +77,7 @@ export default function Compare() {
                   (course.type !== 'combo'
                     ? ' (Overall)' //Indicates that this entry is an aggregate for the entire course/professor
                     : ''),
-                data: Number.isNaN(
-                  categories[0],
-                )
-                  ? grade_dist
-                  : categories,
+                data: Number.isNaN(categories[0]) ? grade_dist : categories,
               };
             })}
           />


src/components/search/SearchBar/SearchBar.tsx
```diff
@@ -286,14 +286,13 @@ export default function SearchBar(props: Props) {
...
+
       if (
-        !(
-          searchQueryLabel(item)
-            .toLowerCase()
-            .includes(newInputValue.toLowerCase()) ||
-          item.subtitle?.toLowerCase().includes(newInputValue.toLowerCase()) ||
-          item.title?.toLowerCase().includes(newInputValue.toLowerCase())
-        )
+        searchQueryLabel(item)
```
egsch: Was removing the `!` here intentional?

PR #572, opened 2026-02-11
Sorting with total students

PyroSh0ck:
## Overview

Added feature to sum up total students per term for a course as well as total students per professor for use in sorting autocomplete results by relevancy.

## What Changed

GenerateAutocompleteGraph.ts was modified substantially, including changes to the addCourse and addProfessor functions to add a totalStudents attribute to the node containing the payload in the radix tree. Likewise, terms are compared based on student count (not recency as this had unpredictable results (see commit 31b163b). The route.ts file for the autocomplete api contains a slight modification that sorts the results from the radix tree in order of student count (with priority given to professors over courses). 

## Other Notes

Many of the courses still do not have totalStudent data, which makes them impossible to sort and they are placed at the bottom. In the future, they should be sorted by the year in which they are meant to be taken (so CS 6V98 should be placed above CS 8V99, which is not the case currently). Likewise, I believe we should be able to get the data for students from elsewhere if its not available from where we are currently retrieving it (possibly coursebook?).

### DIFF
--- a/src/scripts/generateAutocompleteGraph.ts
+++ b/src/scripts/generateAutocompleteGraph.ts
@@ -135,15 +135,6 @@ function addProfessor(
     profFirst: profFirst,
     profLast: profLast,
   };
-  /* 
-    Alternative to data (shallow copy of original prof instead of ref):
-    const data = {
-      ...(originalProf ?? {
-        profFirst: profFirst,
-        profLast: profLast
-      })
-    }
-  */
 
   // saves the node with the data
   const profNode = addWithParents(nodes, profLast, data);


src/app/api/autocomplete/route.ts
```diff
@@ -30,6 +30,22 @@ export async function GET(request: Request) {
...
+      if (aIsCourse) {
+        return -1;
+      }
+      return 1;
+    }
+
+    // sorts results in descending order, while putting any results with 0 students at the end
+    return (b.totalStudents ?? -1) - (a.totalStudents ?? -1);
+  });
+
```
TyHil: Could you move this code inside the `searchAutocomplete` function in `src/modules/autocomplete.ts` so it's not specific to the route?

TyHil (changes requested):
This is super cool. I think a good way to see the impact, just to note for others, is to type "13" into the search bar and see how much more relevant the results are.

I'm gonna wait to review the `generateAutocompleteGraph` file until you go back to summing all the sections instead of taking the max section, as we discussed in dm. So this is just some changes to make to the other files

Issue #573, opened 2026-02-18
[Feature Request] Create Toggle in Search Results To Override Section Type Filter


egsch:
# Overview
In #527 , we're adding a filter for grades of different section types. This filter impacts grade distributions, but we want to add a simple override for this filter that lets user quickly switch to see grades for all sections without changing the filter.
- why do we want this? For example: If John Cole teaches ethics online for the first time next semester, and I set the modality filter to online, I won’t see his grading history, but I would like to know how he has taught in the past in-person before

Inside each card, at the top right, include a horizontal toggle between all, in-person, online, hybrid. This is a local override that **does not change the section type filter**.

# Proposal

- [ ] (Dependent on #527): Add the logic to process this:
   - Override section type in grade calculation when the toggle is changed
   - Automatically set toggle to the option from the filter

PR #580, opened 2026-02-25
made override muitoggle element

### LINKED ISSUE #577: [Feature Request] Search Results Toggle UI
Add a simple toggle component with options "all", "in-person", "online", and "hybrid" inside search result dropdown card that uses the MUI Toggle component. Keep track of which option is selected in (local) state

This toggle will be just above the grade graph for a particular search result or course combo.

See parent #573 - will want to work with assignee to that issue

Snailsy5583:
## Overview
closes #577 #573

Created MUIToggle that is supposed to override the section types filter.

## What Changed

SearchResultsTable.tsx: where the actual override toggle lives, also do the logic to override the chosenSectionTypes
LineGraph.tsx, SingleGradesInfo: had to add an optional parameters in order to change the line graph to correspond with the override toggle

### DIFF
--- a/src/app/dashboard/FilterContext.tsx
+++ b/src/app/dashboard/FilterContext.tsx
@@ -22,6 +22,9 @@ export const FiltersContext = createContext<{
   sectionTypes: string[];
   chosenSectionTypes: string[];
   setChosenSectionTypes: Dispatch<SetStateAction<string[]>>;
+  sectionTypesOverride: string[];
+  chosenSectionTypesOverride: string;
+  setChosenSectionTypesOverride: Dispatch<SetStateAction<string>>;
 }>({
   semesters: [],
   chosenSemesters: [],
@@ -29,6 +32,9 @@ export const FiltersContext = createContext<{
   sectionTypes: [],
   chosenSectionTypes: [],
   setChosenSectionTypes: (_) => _,
+  sectionTypesOverride: [],
+  chosenSectionTypesOverride: "",
+  setChosenSectionTypesOverride: (_) => _,
 });
 
 export default function FiltersProvider({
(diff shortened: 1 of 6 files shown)


egsch:
Hey, thank you for doing this! Sorry this is on me for not fleshing out the description on #577, but could you move this toggle to go inside the individual search result cards? Just above the graph here

image

You'll only need to store the selected option locally inside the card.

Snailsy5583:
Hi,

I'm so sorry, I should have read the issue more clearly. I moved the toggle into the rows, and made sure the state is kept locally. But, I kept all the options still in the FilterContext.tsx file just in case its needed in order to make filter work later. I can remove that as well if needed.

Thank you!

egsch:
> Hi,
> 
> I'm so sorry, I should have read the issue more clearly. I moved the toggle into the rows, and made sure the state is kept locally. But, I kept all the options still in the FilterContext.tsx file just in case its needed in order to make filter work later. I can remove that as well if needed.
> 
> Thank you!

No problem, thank you for making that update! 

I don't think the changes in FiltersContext will be necessary because this will change how the grade info is calculated locally, so you can probably take them out.

I am going to keep this PR open as a draft for right now until we add the functionality rather than merge directly into dev. You're welcome to claim the functionality task #573 if you'd like to finish it (no rush to get it done), or you can switch to work on something else next week

Thanks!

AbhiramTadepalli:
As per yesterday's Trends meeting, we were thinking of turning the toggle into a chip next to Mean GPA

Isoscelestial:
@Snailsy5583 Here's a FilterChip component that you could use: [FilterChip.tsx](https://github.com/UTDNebula/utd-trends/blob/d2b747be410a40093054b998f5f016bcc0bc915f/src/components/search/Filters/FilterChip.tsx). Just copy-paste the file into `src/components/search/Filters/` and it should work for your needs.

The way it works is a little strange, so you should really check out a [usage example here](https://github.com/UTDNebula/utd-trends/blob/d2b747be410a40093054b998f5f016bcc0bc915f/src/components/search/Filters/Chips/MinRatingFilterChip.tsx) (some more examples in [this draft PR](https://github.com/UTDNebula/utd-trends/pull/606/changes))

PR #593, opened 2026-03-15
546 search result fit

### LINKED ISSUE #546: [Feature Request] Indicator When Search Result Fits in Schedule
# Overview

Because search results and section schedule information are on separate pages, it is difficult to tell when available sections fit with elements already selected in your schedule. To remedy this, add an indicator for each combo that is available next semester if there is some section of that combo that fits in the user’s current schedule

# Proposal

In search results, when checking if sections next semester are available for the class, also check whether or not these sections fit in the user’s schedule. If at least one does fit, add:
- A dot/badge on the planner icon, similar to what we have on the “WhatsNew” component
- A modified tool tip – currently “Add to planner”, add something like “Sections fit in your schedule”

OddEyed45:
## Overview

**List the GitHub issues containing the issues relevant to this pull request.**

Resolves #546 

**Describe your changes here at a high level, describing how this PR fits into the
rest of the project.**

Adds an indicator (similar to the one on WhatsNew) on a course's planner icon on search results when it fits the schedule on planner.

## What Changed

**Go more into detail about key files that were modified and why they were
relevant to this feature or bug fix.**

PlannerCard.tsx - changed the hasConflict for export (needed in AddToPlanner.tsx)
AddToPlanner.tsx - first, it gets the available sections from the search result and selects only those for the current semester that is not an exam section. next, it gets the sections already in planner (similar to PlannerCoursesTable.tsx). Then, the overlap is checked for all sections for the current search result. if there are any sections available, there is an indication on the search results page on the planner icon.


src/components/search/SearchResultsTable/AddToPlanner.tsx
```diff
@@ -21,25 +29,131 @@ export default function AddToPlanner({ searchResult }: addToPlannerProps) {
...
+    })
+    .filter((section) => typeof section !== 'undefined');
+
+  const sectionsThatFitPlannerSchedule = availableSections.filter((section) => {
+    return !hasConflict(section, selectedSections); // Keep sections with no conflicts
+  });
+
+  return sectionsThatFitPlannerSchedule.length > 0 ? (
+    <Badge color="secondary" badgeContent=" ">
+      <Tooltip
```
egsch: I think you may be able to avoid repeating all this code by keeping the Badge component all the time and using the [invisible prop](https://mui.com/material-ui/react-badge/#badge-visibility) instead of using a ternary statement

egsch (commented):
Thank you so much @OddEyed45 ! This looks really great! I left a couple of comments in the code but here are also some general questions/comments. Sorry for the late review and all the questions, this is really good!

Visual quibbles:
- Do you think you could try using `variant="dot"` on the badge? Right now it's pretty big and makes the interface feel a bit crowded
- If it's possible, could you try moving the badge wrapper so it directly surrounds the icon without all the padding from the checkbox?

And a little functionality question for anyone who has input: right now the hasConflict function returns `true` for a pair of sections if the conflicting section is the same class as the original section. This is important in Planner, but it does seem a little confusing for the badges in our current implementation; ex. if we have a section selected in Planner it will still show a badge/count for that section saying it fits in your schedule. There's also an issue with labs (ex. it will show a badge for MATH 2418.HN1 when there's an overlapping lab section selected) but that might be a general problem with hasConflict. **Do you think we want badges for all classes where the only conflicts are with other sections of the same class? Is this confusing for anyone else?**

src/components/search/SearchResultsTable/AddToPlanner.tsx
```diff
@@ -75,31 +196,42 @@ export default function AddToPlanner({ searchResult }: addToPlannerProps) {
...
-                effectiveTeachingSemester,
-              );
-            } else {
-              addToPlanner(searchResult.searchQuery, effectiveTeachingSemester);
-              if (addJustCourseToo) {
+        <Badge
+          badgeContent={<EventAvailableIcon sx={{ fontSize: 12 }} />}
+          invisible={finalDisplayCount === 0}
+          overlap="circular"
+          color="secondary"
```
TyHil: This should be `color="primary"` to make it cornflower 300 in dark mode instead of royal.

TyHil (changes requested):
Looks good! I think there's a decision to make tho: Consider the majority of our users that have never added anything to Planner. This badge may help draw attention to the planner button (even tho it's a bit pointless since everything fits their empty schedule) but it also adds more clutter to the dashboard. I'd lean toward the side of only showing the badge if a user already has some section already added to Planner.

src/components/planner/PlannerCoursesTable/PlannerCard.tsx
```diff
@@ -87,7 +87,7 @@ function parseTime(time: string): number {
   return hourNum + minute / 60;
 }
 
-function hasConflict(
+export function hasConflict(
```
AbhiramTadepalli, line 91: I'll have to update the logic in #600 to match this

src/components/search/SearchResultsTable/AddToPlanner.tsx
```diff
@@ -34,16 +38,135 @@ export default function AddToPlanner({ searchResult }: addToPlannerProps) {
...
+      /^[05]/.test(s.section_number) ||
+      /^0[WHL]/.test(s.section_number) ||
+      /^(HON|HN)/.test(s.section_number),
+  );
+  const labSections = courseOnlySections.filter((s) =>
+    /^[1236]/.test(s.section_number),
+  );
+  const examSections = courseOnlySections.filter((s) =>
+    /^7/.test(s.section_number),
+  );
```
AbhiramTadepalli: For all 3 of these, do we want to move the logic into `SearchQuery.tsx` next to `sectionCanOverlap`?

src/components/search/SearchResultsTable/AddToPlanner.tsx
```diff
@@ -34,16 +38,135 @@ export default function AddToPlanner({ searchResult }: addToPlannerProps) {
...
+  const latestSections = allResults.map((r) =>
+    r.isSuccess
+      ? r.data.sections.filter(
+          (s) =>
+            s.academic_session.name === effectiveTeachingSemester && s != null,
+        )
+      : [],
+  );
+
+  const selectedSections = planner
```
AbhiramTadepalli, line 69: We do this (get allResults & latestSections) 2+ times in our code, should we move it into a module / export a function from PlannerCoursesTable or smtg that does it in both places? Or not needed @TyHil

src/components/search/SearchResultsTable/AddToPlanner.tsx
```diff
@@ -34,16 +38,135 @@ export default function AddToPlanner({ searchResult }: addToPlannerProps) {
...
+        const courseHasLabs = labSections.length > 0;
+        const courseHasExams = examSections.length > 0;
+
+        if (fitSections([lecture]).length === 0) {
+          return false;
+        }
+
+        if (courseHasLabs) {
+          const matchingLab = labSections.find(
+            (l) => getBaseSectionNumber(l.section_number) === base,
```
AbhiramTadepalli: correct me if I'm wrong @OddEyed45 but is this line only comparing the last 2 digits of the section number? What's the reasoning for that?

src/components/search/SearchResultsTable/AddToPlanner.tsx
```diff
@@ -75,31 +201,42 @@ export default function AddToPlanner({ searchResult }: addToPlannerProps) {
...
+              e.stopPropagation(); // prevents opening/closing the card when clicking on the compare checkbox
+              if (inPlanner) {
+                removeFromPlanner(
+                  searchResult.searchQuery,
+                  effectiveTeachingSemester,
+                );
+              } else {
                 addToPlanner(
-                  convertToCourseOnly(searchResult.searchQuery),
+                  searchResult.searchQuery,
```
AbhiramTadepalli, line 232: @egsch I might have missed this in the other PR but, just for my understanding, why was this changed from `convertToCourseOnly(searchResult.searchQuery),`?

egsch: This is just the GitHub changes showing weird, you can see the equivalent code on 88 in the old version (and the one with convertToCourseOnly is on 226 here)

src/components/search/SearchResultsTable/AddToPlanner.tsx
egsch: Yeah I think that's a good idea, I'll do that

src/components/search/SearchResultsTable/AddToPlanner.tsx
OddEyed45: I messed up that part. I pushed fixes for this and tested it with MATH 2418 and CS 4141.

OddEyed45: I only changed the part from declaring labSections, lectureSections... onward to the final line where finalDisplayCount is changed. All of your + Emma's changes are kept intact.

src/components/search/SearchResultsTable/AddToPlanner.tsx
AbhiramTadepalli: [ee73be4](https://github.com/UTDNebula/utd-trends/pull/593/commits/ee73be4c3c70044056e6db533e7b5901d2a39385)

src/components/search/SearchResultsTable/AddToPlanner.tsx
AbhiramTadepalli: I [updated](https://github.com/UTDNebula/utd-trends/pull/593/commits/5249557f6ccc78dc1b2ce5ad09270e8ecd2c311a) one clause in your logic but otherwise looks good!

TyHil (approved):
🎉🎉🎉

src/components/planner/PlannerCoursesTable/PlannerCard.tsx
egsch: Resolving this comment so we have everything resolved as we can merge this in

egsch (approved):

PR #597, opened 2026-03-25
implement aggregate course searching

### LINKED ISSUE #564: [Feature Request] Allow searching incomplete search terms for groups of classes (ex. CS 13XX)
# Overview

If a person wants to search for multiple courses with commonalities in course number, they have to add all these courses separately.
Required features from discussion last semester:
 - Should show up as soon as you type “CS 43“
 - Should be available for all ##XX
 - Should have a title (like course title) like “CS courses starting with 43“
 - Add a dashed underline on the Overall result options to highlight the existence of hover
 - Show overalls and professor combos for 43xx

# Proposal

- [ ] Create a way to process searching for these combinations of courses and combine the results in search results
- [ ] Add new entries to autocomplete for these combinations
- [ ] Update UI as required to meet requirements above

jakobqnen:
## Overview

Resolves #564  

This change allows for searching on incomplete course numbers, such as 'CS13XX' or 'MATH2XXX' . When these aggregates are searched, the results are the Overalls for each course matching the query. So searching 'CS13XX' results in the Overalls for CS1336, CS1337, etc. being displayed.

## What Changed

Created a new table to store the courses for each aggregate, course_aggregates_table.json . This table is queried to find when to create the aggregate option on the SearchBar (SearchBar.tsx). It is also queried when fetching query results (createSearchQuery.tsx). Instead of adding the Overall query and all combo queries to get the results of, we instead get all the Overall queries for the courses that fall under the aggregate search. 

## Other Notes

Normally, the aggregate option should appear between the matched recents and the rest of the matching results in the autocorrect dropdown. There may be cases where this isn't the most logical place for it.


TyHil (commented):
This functionality looks good! The new table is 111Kb so idk about loading it on the client side in search bar. We could potentially load it in the autocomplete endpoint and return smth like:
```
{
  "newAggregrateData": ...,
  "autocomplete": ...,
}
```

Also I'm not sure if this is intentional but the functionality that automatically adds a tag chip to the search bar when you hit space doesn't work for these aggregates. Like type "CS 1200 " and it turns into a tag. But type "CS 23XX " and it doesn't.

egsch (approved):
Sorry for the delay @jakobqnen but thank you so much, this looks great! Excited to add this!

src/app/dashboard/page.tsx
```diff
@@ -182,7 +182,9 @@ export default async function Page({ searchParams }: Props) {
                     }
                   >
                     <Right
-                      courses={courses}
+                      courses={courses.filter(
+                        (course) => !course.number?.includes('X'), // Don't try to show overview for aggregate course searches
+                      )}
```
egsch, line 189: Hey I think this change is only applying on mobile right now, can you add it to the desktop layout a few lines above as well?

jakobqnen: @egsch thanks for pointing that out, I've fixed that in the newest commit

Issue #576, opened 2026-02-24
[Feature Request] Mobile UI: Rotate compare bar graph


egsch:
# Overview
If more than one class is in Compare on mobile, the compare bar chart's x-axis and legend become very crowded.

<img width="510" height="502" alt="Image" src="https://github.com/user-attachments/assets/c33fbb43-30b5-44f1-98f5-096fde278e05" />

# Proposal

In mobile, rotate the graph to match the design below as closely as possible.
 -  Since it's not specified in the design, test moving the bar/line graph toggle to the top of the graph. and see how it looks

## Supplementary Information
<img width="328" height="710" alt="Image" src="https://github.com/user-attachments/assets/98a924b1-0950-45c6-ad52-deca610b9bf7" />

PR #599, opened 2026-03-28
feat: rotate bar graph to horizontal on mobile for compare page (#576)

MuhammadBilal350:
rotated the comparision bar graph horizontally for the grades of professors ,only for mobile

### DIFF
--- a/src/components/compare/Compare/Compare.tsx
+++ b/src/components/compare/Compare/Compare.tsx
@@ -25,6 +25,14 @@ export default function Compare() {
     return <p>Click a checkbox to add something to compare.</p>;
   }
 
+  // Only chart courses that actually have grade data for the current filters
+  const compareWithData = compare.filter((course) => {
+    const categories = convertNumbersToPercents(
+      calculateGrades(course.grades, chosenSemesters, chosenSectionTypes),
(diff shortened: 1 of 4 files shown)


egsch:
Hey @MuhammadBilal350 super sorry for the late response! From our meeting last week, we want to make sure that this graph is turned on mobile in the compare page, but not in search result cards- is that change something you would be able to finish up?
If you could merge develop into your branch as well that would be great! After you do that, feel free to use the mobile app and see if there are any small tweaks ex. adjusting table height, padding you would like to make.

MuhammadBilal350:
Hi @egsch, i can make the change of the placement such that the bar graph is showing in the compare section and not the search cards. Thank you for notifying will get it done before next meeting

egsch:
> Hi @egsch, i can make the change of the placement such that the bar graph is showing in the compare section and not the search cards. Thank you for notifying will get it done before next meeting

Thank you! 

Just realized my wording may have been a little unclear so just to clarify: the graph should be visible in both spots but should only be rotated in the compare table and should remain horizontal on the search results.

MuhammadBilal350:
Sounds good @egsch

PR #600, opened 2026-03-29
Auto gen schedule

AbhiramTadepalli:
## Overview

Auto-complete a myplanner schedule

## What Changed

It uses the same `hasConflict()` logic we already have alongside a constraint-sat algo to try and select sections from each course in MyPlanner. If a user has already selected a few, it respects the decision and works around it

## Other Notes

Tried to implement a greedy-ish approach that is more likely to select
- sections with professors assigned
- sections with professors who are higher-rated in rmp
- sections of a professor with a higher median-gpa
- sections that start after 9am and before 5pm

### DIFF
--- a/src/modules/scheduleGenerator.ts
+++ b/src/modules/scheduleGenerator.ts
@@ -65,7 +65,7 @@ export function checkConflict(
   return false;
 }
 
-// heuristic to core a section based on its desirability
+// heuristic to score a section based on its desirability
 function scoreSection(
   section: Section,
   meetings: ParsedMeeting[],


src/components/planner/PlannerCoursesTable/PlannerCard.tsx
```diff
@@ -75,65 +76,40 @@ export function LoadingPlannerCard() {
...
-                newSection.course_details[0].subject_prefix &&
-              selectedSection.course_details[0].course_number ==
-                newSection.course_details[0].course_number
-            )
-              return false; // if times overlap, but same course, we can switch it out without conflict
-            return true; // Conflict detected
-          }
-        }
-      }
+    const newSectionMeetings = newSection.meetings
```
TyHil: Has the output of `hasConflict` changed? Or just the way in which it computes its output? Like is the new function equivalent to the old?

src/components/planner/PlannerCoursesTable/PlannerCoursesTable.tsx
```diff
@@ -5,14 +5,18 @@ import PlannerCard, {
...
 import {
   convertToCourseOnly,
   removeSection,
   searchQueryLabel,
   searchQueryMultiSectionSplit,
+  type SearchQuery,
 } from '@/types/SearchQuery';
-import { Alert, Snackbar, Typography } from '@mui/material';
+import AutoFixHighIcon from '@mui/icons-material/AutoFixHigh'; // Added Icon
+import { Alert, Button, Snackbar, Typography } from '@mui/material'; // Added Button
```
TyHil: AI comments 😔

Remove

src/modules/scheduleGenerator.ts
```diff
@@ -0,0 +1,280 @@
...
+        sec1Meeting.end > sec2Meeting.start
+      ) {
+        return true; // Conflict detected
+      }
+    }
+  }
+  return false;
+}
+
+// heuristic to core a section based on its desirability
```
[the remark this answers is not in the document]
AbhiramTadepalli: wdym? I core sections all the time

PR #618, opened 2026-04-18
Search other professors in planner

### LINKED ISSUE #440: [Feature Request] Allow searching from within a MyPlanner card
# Overview

If users want to see a different professor for the same class, add a button (probably inside that card) that searches for that course.

## Supplementary Information
Use the SearchTerm's prefix and number props to build the search. Ig you can navigate to that url

****
No longer on hold! Additional specifications: 

Add a context menu (dropdown) that replaces the current "remove from planner" button. For now, keep the same icon. On click, show a menu with two options:
   - option to remove from planner
   - option to go to the search results for that course

Second option will be a link to the search results page for this course (add params for course prefix + number)

mikehquan19:
## Overview

Resolves #440,

Add the new footer at the bottom of the section planner card that allow for searching for other professors of the same given course. 

Screenshot 2026-04-17 at 20 20 00

## What Changed
Add the footer at the bottom of list of available sections when the user expands. When the user clicks on the button, it will navigate to the page listing other professors of the course. 

My concern is that I'm not sure if this should be a button or a link, so some feedback would be great.

### DIFF
--- a/src/components/planner/PlannerCoursesTable/PlannerCard.tsx
+++ b/src/components/planner/PlannerCoursesTable/PlannerCard.tsx
@@ -22,6 +22,7 @@ import BarChartIcon from '@mui/icons-material/BarChart';
 import BookIcon from '@mui/icons-material/Book';
 import BookOutlinedIcon from '@mui/icons-material/BookOutlined';
 import EventIcon from '@mui/icons-material/Event';
+import SearchIcon from '@mui/icons-material/Search';
 import KeyboardArrowIcon from '@mui/icons-material/KeyboardArrowRight';
 import {
   Box,
@@ -383,6 +384,7 @@ function SectionTableRow(props: SectionTableRowProps) {
 function SearchOtherProfessorFooter(props: {
   course: Course | Record<string, never>;
   teachingSemester: string;
+  bottomBorder: boolean;
 }) {
   const params = new URLSearchParams();
   if ('prefix' in props.course && 'number' in props.course) {
@@ -398,11 +400,19 @@ function SearchOtherProfessorFooter(props: {
     <TableRow>
       <TableCell
         colSpan={6}
-        className="py-1 border-t-1 border-t-royal dark:border-t-cornflower-300"
+        className={`
+          py-2 border-t border-t-royal dark:border-t-cornflower-300
+          ${
+            props.bottomBorder
+              ? 'border-b border-b-royal dark:border-b-cornflower-300'
+              : 'border-b-0'
+          }
+        `}
         align="center"
       >
         <Button
-          className="normal-case"
+          className="normal-case rounded-full"
+          variant='contained'
           LinkComponent={Link}
           href={`dashboard?${params.toString()}`}
           onClick={() => {
@@ -410,6 +420,7 @@ function SearchOtherProfessorFooter(props: {
             // instead of clicking "Search Results"
             sessionStorage.setItem('dashboardSearchTerms', params.toString());
           }}
+          startIcon={<SearchIcon />}
         >
           See other professors
         </Button>
@@ -789,6 +800,10 @@ export default function PlannerCard(props: PlannerCardProps) {
                 <SearchOtherProfessorFooter
                   course={convertToCourseOnly(props.query)}
                   teachingSemester={props.teachingSemester}
+                  bottomBorder={
+                    latestExtraSections !== null &&
+                    latestExtraSections.sections.length > 0
+                  }
                 />
               </TableFooter>
             </Table>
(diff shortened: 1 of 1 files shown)


TyHil:
I think a link is best so Next can preload the page when the link is hovered. You'll want to do it like explained in this issue: https://github.com/utdnebula/utd-clubs/issues/532.

Also maybe centered and with `className="normal-case"`

mikehquan19:
> I think a link is best so Next can preload the page when the link is hovered. You'll want to do it like explained in this issue: [UTDNebula/utd-clubs#532](https://github.com/UTDNebula/utd-clubs/issues/532).
> 
> Also maybe centered and with `className="normal-case"`

Ok sweet, fix asap

egsch:
Hey thank you so much for adding this!

I feel bad now bc I just realized I may not have edited the original description in #440 when I added extra specifications at the end of the issue. I think the original intent was to add the option to add to planner to a  new drop-down menu on the Add To Planner button  rather than putting it in a footer. That being said , I do like the footer and I'm totally up to keep the footer if y'all want to! 

Super sorry for not clarifying this and lmk what youd'd like to do there!

mikehquan19:
> Hey thank you so much for adding this!
> 
> I feel bad now bc I just realized I may not have edited the original description in #440 when I added extra specifications at the end of the issue. I think the original intent was to add the option to add to planner to a new drop-down menu on the Add To Planner button rather than putting it in a footer. That being said , I do like the footer and I'm totally up to keep the footer if y'all want to!
> 
> Super sorry for not clarifying this and lmk what youd'd like to do there!

Oh yeah, I was going to implement because I think looks cool too, but Abhiram said he'd be mad if I implemented it and he suggested the footer haha.

mikehquan19:
Screenshot 2026-04-17 at 21 06 08

Now it looks something like this, but there's still the button-like effect when we hover over. It's expected tho since this is still Button with LinkComponent.

mikehquan19:
Screenshot 2026-04-17 at 21 28 32

Would we want to have a little border like this ?

AbhiramTadepalli:
@egsch I was thinking impeding users from quickly removing from planner would be bad UI (one click vs two clicks). I rather like the footer to the sections

egsch:
> @egsch I was thinking impeding users from quickly removing from planner would be bad UI (one click vs two clicks). I rather like the footer to the sections

Yeah this makes sense! I think the intention of the change as we discussed it before was partially to prevent users from accidentally removing selected classes from planner + to avoid clutter on the cards, but yeah I think we could leave the footer (and maybe disable removing for planner or add something there only if they have a section of that course selected?).

mikehquan19:
@egsch Can you elaborate a bit on the remove button disable? I will surely be able to implement it but I just need a clearer idea.

egsch:
> @egsch Can you elaborate a bit on the remove button disable? I will surely be able to implement it but I just need a clearer idea.

Hey sorry that was just an idea I was throwing out there and kinda a separate thing, idk if we want to do it so no need to implement

On this PR, do you think we would want to hide the footer on overall classes since there aren't really "other professors" to search for? Or is it better to keep the footer just in case they still want to add other combos?

mikehquan19:
> > @egsch Can you elaborate a bit on the remove button disable? I will surely be able to implement it but I just need a clearer idea.
> 
> Hey sorry that was just an idea I was throwing out there and kinda a separate thing, idk if we want to do it so no need to implement
> 
> On this PR, do you think we would want to hide the footer on overall classes since there aren't really "other professors" to search for? Or is it better to keep the footer just in case they still want to add other combos?

If it’s an overall class, can we still click that button to search for all the professors? I’m inclined to hide it tho because you are right there are no other professors, but wdym by adding other combos?

AbhiramTadepalli:
I think we should keep the footer for overalls. Because I may want to see who my options are for that course.

@mikehquan19 combos are course+professor `CS 1200 John Cole`, whereas an overall is just a course `CS 1200`

egsch:
Hey sorry for the delay on this! It looks great and yeah I think keeping the button on combos is fine if y'all like. I did just notice that the border is a little funky when we have extra exam/lab sections, do we want to add a border on the bottom in this case? Or move the footer below the extra section card?

image

AbhiramTadepalli:
I'm really unsure which would work better. Leaving it as is with better dividers would make sense because it comes right after that prof's sections, but bottom of the card might be more standard. I'm leaning towards the former if we can make it look good.

@mikehquan19 could you
- [ ] add a border to the bottom of "see other professors"
- [ ] make a button that encompasses the text, so users know it's click-able and/or add the search icon to the left of the text

mikehquan19:
Ok, I’ll do it. And make a button to signify clickability is good idea

mikehquan19:
Screenshot 2026-05-01 at 00 24 07

PR #620, opened 2026-04-18
feat: sort course name autocomplete results by totalStudents

### LINKED ISSUE #517: [Feature Request] Improve homepage search result
# Overview

Right now, the search bar result shows **EVERYTHING** (courses and prof names), and we had to limit the shown result. This results in user having to provide additional keyword to search for what they wanted.

# Proposal

Categorize the search result into undergrad courses, grad courses, and prof name (idea from RMP) to be able to display all the result that the search algorithm generates and tailors more to our target audience (undergrads).

To do this, we are weighing the course priority (1000, 2000, 3000, 4000) in the search algorithm.

## Supplementary Information
Eg: I want to search for Introduction to Machine Learning (CS 4375), but the search alg does not prioritize that result unless provided more detail, like "CS".

Image
 

Image

SinhSinhAn:
## Summary

When a user searches by course name (e.g. \"machine learning\" instead of \"CS 4375\"), the autocomplete results were ranked purely by text relevance. This meant that obscure or graduate-level courses could appear above popular undergrad ones even when they were equally relevant matches for example, CS 6375 showing up before CS 4375.

This PR fixes that by using enrollment data as a tiebreaker. Here is what changed:

**`src/scripts/generateCourseNameTable.ts`** The script that builds `course_name_table.json` now computes a `totalStudents` value for each course entry by summing `total_students` across every section across every academic session in the aggregated data. This follows the same pattern already used in `generateAutocompleteGraph.ts` from PR #572.

**`src/app/api/courseNameAutocomplete/route.ts`**  After the existing relevance filter (distance + standard deviation cutoff), results are now sorted by `totalStudents` descending before being returned. Courses that more students have taken will rank above equally-relevant but less popular ones.

Note: `course_name_table.json` needs to be regenerated via `npm run buildcoursenames` for the `totalStudents` values to appear in the output.

Closes #517


src/app/api/courseNameAutocomplete/route.ts
```diff
@@ -237,6 +239,9 @@ export async function GET(request: Request) {
   const oneStdCutoff = cut + 1 * stdDev; // 1 standard deviation cutoff
   const resultsWithoutDistance: Result[] = results
     .filter((r) => r.distance <= oneStdCutoff)
+    .sort(
+      (a, b) => (b.result.totalStudents ?? -1) - (a.result.totalStudents ?? -1),
```
AbhiramTadepalli: why not default to 0 if totalStudents is null?
This has an edge case where `b.result.totalStudents` == 1 & `a.result.totalStudents` is null -> `1 - (-1)` in the sorting which considers b equal to a. If you default to 0 this treats classes with null totalStudents as classes that haven't been taught before

src/app/planner/Planner.tsx
```diff
@@ -55,12 +56,16 @@ export default function Planner() {
...
-        )
+        <div className="flex flex-col">
+          {planner.filter(
+            (entry) => entry.semester === effectiveTeachingSemester,
+          ).length === 0 ? (
+            <PlannerEmpty />
+          ) : (
+            <PlannerCoursesTable />
+          )}
+          <ClubsBanner />
```
AbhiramTadepalli: Can you remove the `<ClubsBanner/>` code?

src/scripts/generateCourseNameTable.ts
```diff
@@ -28,10 +34,26 @@ for (let prefixItr = 0; prefixItr < aggregatedData.data.length; prefixItr++) {
     courseNumberItr++
   ) {
     const courseNumberData = prefixData.course_numbers[courseNumberItr];
+    let courseStudents = 0;
+    for (
+      let sessionItr = 0;
+      sessionItr < courseNumberData.academic_sessions.length;
```
AbhiramTadepalli, line 40: Instead of a for loop, replace the outer code with `courseNumberData.academic_sessions.map((sessionData) => {}`

src/scripts/generateCourseNameTable.ts
```diff
@@ -28,10 +34,26 @@ for (let prefixItr = 0; prefixItr < aggregatedData.data.length; prefixItr++) {
...
+      sessionItr++
+    ) {
+      const sessionData = courseNumberData.academic_sessions[sessionItr];
+      for (
+        let sectionItr = 0;
+        sectionItr < sessionData.sections.length;
+        sectionItr++
```
AbhiramTadepalli, line 47: and likewise do `sessionData.sections.map((section) => {}`

src/scripts/generateCourseNameTable.ts
AbhiramTadepalli: actually, let me take a stab at it using a reduce

src/app/api/courseNameAutocomplete/route.ts
```diff
@@ -237,6 +239,9 @@ export async function GET(request: Request) {
...
+    )
     .map((result) => ({
       title: result.title,
       result: result.result,
```
AbhiramTadepalli, line 248: @egsch do we want to display how many students took that course in the frontend? If we do it for search-by-course-name we should probably do it for professors and search-by-course-number too, which we can spinoff into a diff issue

AbhiramTadepalli (changes requested):

src/scripts/generateCourseNameTable.ts
AbhiramTadepalli: @TyHil is it better to define an interface for the `aggregated_data.json` data so that we can reduce in typescript, or leave it as for loops?

AbhiramTadepalli:
Question: should we include totalStudents as a weighted parameter in the algo as well? Searching for "Machine Learning" doesn't show CS 4375 because it misses the cutoff ("Introduction to Machine Learning" has a bigger edit distance than, say "Statistical Machine Learning")

src/scripts/generateCourseNameTable.ts
TyHil: Sounds like pretty equivalent solutions tbh. Idk maybe types are better but the for loops work and is what we do in the other script files

TyHil:
I think the issue was to use totalStudents as a weighted parameter. Since the results of the course name search are already sorted by the algorithms determined similarity, we don't need to override that and sort by totalStudents but instead factor that into the existing sorting

AbhiramTadepalli (approved):
LGTM

TyHil:
Once this is merged, would whoever merges it add a quick issue to Notebook to replicate this functionality there? Just link to this PR and say the issue is waiting on https://github.com/UTDNebula/utd-notebook/pull/173 if that PR is not already merged.

PR #621, opened 2026-04-19
added up/down indicator search results.

### LINKED ISSUE #252: [Feature Request] Increasing/decresing GPA indicator
# Overview

This idea may need some refinement but ideally there should be a way to tell how the GPA for a search result has been changing across semesters. Without having to check through each semester in the filter.

# Proposal

This could be a graph of the GPA over time or just an [trending up/down icon](https://mui.com/material-ui/material-icons/?selected=TrendingDown) based on some math to tell if the GPA has been increasing.

- [ ] Logic to detect GPA trend in recent semesters
- [ ] Display it (somehow tbd)

## Supplementary Information

Since this will be a greater use of semester by semester data, it would be good to indicate from how many semesters each results data is coming from, maybe just another item below the graph on each down that says "Semesters: x"

greeshiee:
## Overview

Resolves #252.

This PR adds a trend indicator next to Mean GPA in results, showing whether performance in the 4 most recent long semesters (Fall/Spring only) is trending up or down compared to all prior long semesters. The comparison uses the unweighted average of per-semester mean GPAs and only appears when there’s enough recent data (grades in at least 2 of the last 4 long semesters). No icon is shown in case of no change.

## What Changed

src/modules/gpaTrend.ts: Introduced a small helper that computes the trend direction (up / down / null) by:
selecting the 4 most recent long semesters,
requiring at least 2 usable recent semesters,
comparing recent vs prior per-semester mean GPAs (ignoring Summer).
src/components/common/SingleGradesInfo/SingleGradesInfo.tsx: Updated the Mean GPA display to render Material UI TrendingUpSharp / TrendingDownSharp icons inline when the helper reports a trend. The computation respects the currently selected section types.


AbhiramTadepalli:
CS 1200 overall shows a downward trend, but it's just .01 off. Can we have a small threshold?
> Check out Kangkook jee's 6332, why is it an upward trend?
Okay so I realized that it's comparing recent sems to what came before, right? It does seem misleading especially for a class like this one because i visually see a downward trend in the line graph recently

greeshiee:
I can place a threshold of 0.025 - 0.05 or so. As for CS 6332, I'm unsure of how we can indicate correctly since the difference is quite significant (prev avg: 3.428, recent avg: 3.564, diff: 0.136). Is there any particular change in my method that would be more representative of what we're indicating?

AbhiramTadepalli:
Did we have a reason against calculating the slope over the last 4 long semesters?

greeshiee:
We decided against using slope for trend calculation since some commented it felt misleading when the line graph as a whole trends one way and a subset of data trends another way and indicating for the subset is misinterpreted when the entire line graph is available for review. Fine tuning slope thresholds was also a challenge that set that to the side.

PR #631, opened 2026-05-11
Update CODEOWNERS

NishilJ:
## Overview

Update CODEOWNERS relevant to org structure. Add API Lead for API-related folders.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,8 +1,8 @@
 # Request Portfolio Lead and Head of Engineering
 * @UTDNebula/portfolio-lead @UTDNebula/head-of-engineering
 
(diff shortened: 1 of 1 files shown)


TyHil:
Idt the API lead needs to be on any of the Trends folders

NishilJ:
> Idt the API lead needs to be on any of the Trends folders

I was also thinking in the context of Mike going to be "senior engineer" but I can understand wym.

TyHil:
Since his senior engineering role is separate from his role as API lead it would be better to @ his GitHub username instead of the team then

TyHil (approved):

AbhiramTadepalli (approved):

PR #613, opened 2026-04-07
Darkmode color changes 

JoelZuniga:
## Overview

Resolves #558

Standardizing dark mode and light mode colors to match org wide presets. 

## What Changed

Added colors for light and dark mode, changed main background and card backgrounds for visibility

Issue #558, opened 2026-01-17
[Feature Request] Update dark mode page and card colors


TyHil:
The Clubs team has had discussion on the background and page colors of the Clubs site, which were originally pulled from Trends. A perfectly black background is bad practice and a non-tinted card color aligns better with dark mode.

Add the following colors to `globals.css`:
```
--color-cornflower-950: #141327;
--color-light: #f6f6f6;
--color-dark: #101010;
```

Update the `body` `className` in `layout.tsx`:
```
// Example from change in Clubs
<body
  className={`bg-light dark:bg-dark ${inter.variable} font-main ${baiJamjuree.variable} text-haiti dark:text-white`}
>
```

Update the backgrounds of any cards to be `bg-white dark:bg-neutral-800`.

This is a cross-project issue so making the same changes to Trends, Skedge, Rooms, Notebook, and Clubs on a branch in each project is required.

See:
https://github.com/UTDNebula/utd-rooms/issues/99
https://github.com/UTDNebula/utd-notebook/issues/61
https://github.com/UTDNebula/skedge/issues/102
https://github.com/UTDNebula/utd-clubs/issues/557

Write access to Nebula projects is managed on a per-team basis so you may have it for one project but not the others. Dm on Discord if you need access for one of the other projects.

PR #635, opened 2026-06-04
Separate RMP and syllabus summaries

### LINKED ISSUE #607: [Bug]: Syllabus Summary not Showing for Professors without RMP

The Syllabus summary should show if the professor has a latest syllabus, even if they have no RMP page or grades.

### What happened?

In our current implementation from #551, if a professor is not in RateMyProfessor, their syllabus summary won't be show as <SingleProfInfo/> isn't rendered.

We should fix this by separating the SingleProfInfo component into two separate components, so the syllabus summary is in its own component and not affected by RMP results. If you need to adjust the UI (which currently has the RMP link and RMP dropdown on a single line) slightly to do this that's OK, just get approval

### Version

trends.utdnebula.com

_No response_

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

TyHil:
Separated the the RMP and syllabus summary features in card dropdowns so now you can see the syllabus summary when the professor does not have any RMP. I also updated the style to use dense MUI tables so it's more recognizable as multiple tables.

In the process, I needed the BaseCard component added to all the projects in the #558 set of issues and realized it should have been added in #613. I missed that in the review so added it here. This matches card colors, rounded corners, and shadows with the rest of the Nebula projects. For the most part the cards just look a tad lighter, easy to see the difference by tabbing between https://dev.trends.utdnebula.com/dashboard?searchTerms=John+Cole&availability=26F and https://utd-trends-git-607-separate-rmp-and-syllabus-utdnebula.vercel.app/dashboard?searchTerms=John+Cole&availability=26F.

Before:
image

After:
image
(expanded)
image

### DIFF
--- a/src/components/common/SyllabusSummary/SyllabusSummary.tsx
+++ b/src/components/common/SyllabusSummary/SyllabusSummary.tsx
@@ -1,7 +1,20 @@
 'use client';
 
+import BaseCard from '@/components/common/BaseCard/BaseCard';
 import { type SearchQuery } from '@/types/SearchQuery';
-import { Collapse, Link, Skeleton, Typography } from '@mui/material';
+import {
+  Grid,
+  Skeleton,
+  Table,
+  TableBody,
+  TableCell,
+  TableContainer,
+  TableHead,
+  TableRow,
+  Tooltip,
+  Typography,
+} from '@mui/material';
+import Link from 'next/link';
 import React, { useEffect, useState } from 'react';
 
 export function LoadingSyllabusSummary() {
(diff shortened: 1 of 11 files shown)


src/components/search/SearchResultsTable/SearchResultsTable.tsx
TyHil, line 1: Note: all of the changes in here after line 500 are only indentation, except for removing a couple usages of `getCellSx` and replacing them with a `className` since `getCellSx` wasn't working.

src/components/common/SyllabusSummary/SyllabusSummary.tsx
```diff
@@ -1,25 +1,171 @@
...
+      {/* Grade Scale Table */}
+      <Grid size={6}>
+        <BaseCard className="dark:bg-neutral-700">
+          <TableContainer>
+            <Table size="small" aria-label="grade scale table">
+              <colgroup>
+                <col className="w-1/2" />
+                <col className="w-1/2" />
+              </colgroup>
+              <TableHead>
```
AbhiramTadepalli: wow :)

AbhiramTadepalli: this looks so nice

AbhiramTadepalli (changes requested):
Looks really good! Do you think we can add a divider, header, or some spacing between the rmp and syllabus cards? It sort of feels like the weighting and grade scale go with RMP.

Also in light mode, can you darken the weighting and grade scale cards? Or give it a light cornflower bg color

TyHil:
Thank you for the good ideas and points, I went with `bg-neutral-100` tables on light mode and MUI's `Divider`s. Lmk what you think

image

AbhiramTadepalli (approved):
LGTM

NishilJ (commented):
Nice job, looks good, but was wondering if it would make sense to extract a component for the weighting and grade tables and their loading variants. Could reduce a decent amount of duplicated code.

TyHil:
Good catch, didn't even think of how much code duplication I was doing. `SyllabusSummary` is now over 100 lines shorter

NishilJ (approved):
thanks

PR #636, opened 2026-06-07
Add footer

### LINKED ISSUE #355: [Feature Request] Indicate how up-to-date every source of data is
# Overview

Trends uses multiple sources of data, but each one has a different level of recency and time ranges

# Proposal
- [ ] Identify all the sources of data we use
- [ ] Print the date range of each data source to the console
- [ ] Indicate the recency of the data in the website (need to wait for Design)

TyHil:
Resolves #355 in that one can open each source of data to see how recent it is.

Basically the same footer as Rooms

image


mikehquan19:
This is not about this trends footer, but should we add the projects column for Rooms footer?

TyHil:
@mikehquan19 see https://github.com/UTDNebula/utd-rooms/pull/123

PR #638, opened 2026-06-21
Update packages

TyHil:
Based on https://github.com/UTDNebula/utd-clubs/pull/698 and https://github.com/UTDNebula/utd-clubs/pull/704

## Changes
- update to Typescript 6
  - remove deprecated stuff from `tsconfig.json`
- update most packages to their latest versions
- add `type:check` npm script
- update `lint-check` GitHub workflow to also validate types

## Notes
- We cannot update ESLint to version 10.0 because [`eslint-plugin-jsx-a11y` does not support it](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/1075). However, that package hasn't been updated in 1.5 years, so we may need to consider either removing it or finding a replacement.
- [ts-node](https://www.npmjs.com/package/ts-node) hasn't been updated in several years, do we need to switch to [tsx](https://www.npmjs.com/package/tsx)?

## After Merge

Remove the "Run ESLint + Prettier" check and add the "Run ESLint + Prettier + TypeScript" check in both the develop and main branch protection rules under "Require status checks to pass before merging" at https://github.com/UTDNebula/utd-trends/settings/branches.

### Before

55 vulnerabilities (1 low, 42 moderate, 10 high, 2 critical)

### After

51 vulnerabilities (1 low, 39 moderate, 9 high, 2 critical)


NishilJ:
looks good, I think we should migrate to tsx

TyHil:
Cool, how about I add switching to `tsx` to the description of https://github.com/UTDNebula/nebula-library/issues/11 so it can all be done at once (for Trends and Notebook). Then this can merge

NishilJ:
sure

NishilJ (approved):

PR #646, opened 2026-08-29
fix(course-overview): preserve periodless descriptions

ColumbusLabs:
## Overview

Course descriptions without a terminal period no longer collapse to an empty string. The parser now removes trailing contact-hours/frequency metadata only when it follows an actual terminal period; otherwise it preserves the complete description, including internal periods such as `U.S.`.

## What Changed

- updated `parseDescription` to recognize the catalog metadata suffix before trimming at the last period
- exported the parser for focused unit coverage
- added cases for the reported period-less LIT 1301 description, a period-less description with an internal period, and the existing terminal-period behavior

## Verification

- `npm test -- --runInBand` — 2 suites, 5 tests passed
- `npm run lint:check`
- `npm run format:check`
- `npm run type:check`

All commands ran with Node 22 as required by the repository.

### DIFF
--- a/__tests__/CourseOverview.test.ts
+++ b/__tests__/CourseOverview.test.ts
@@ -1,0 +1,45 @@
+import { parseDescription } from '@/components/overview/CourseOverview/CourseOverview';
+import type { Course } from '@/modules/fetchCourse';
+
+function createCourse(description: string): Course {
+  return {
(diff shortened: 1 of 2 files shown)


AbhiramTadepalli:
@justinschwerdtfeger this might be a bot?

